PR #3, opened 2016-01-04
Put best practice in place to use "save exact" for npm packages.

nickytonline:
@coryhouse, here's a suggestion to put the best practice in place to use save-exact=true for npm packages.


coryhouse:
Thanks Nick. Great call!

DannyDelott:
@coryhouse Awesome starter kit and documentation.
@nickytonline I like this suggestion. Could you elaborate why this is best practice?

nickytonline:
@DannyDelott please see issue #4 for links explaining this.

PR #7, opened 2016-01-05
Fix Browsersync URL and name trademark spelling

demonkoryu:


coryhouse:
Good catch. Thanks for the PR!

Issue #1, opened 2016-01-02
Run tests as part of build step


coryhouse:
Right now, you have to hit Ctrl+c/Cmd+c to kill the process since the test script keeps watching files after the build completes. No biggie, but a minor annoyance worth fixing.

nickytonline:
@coryhouse I don't see this happening when I build. Here's my output and then the build completes.

```
unicorn:react-slingshot nickytonline$ npm run build

> react-slingshot@1.0.1 prebuild /Users/nickytonline/GitHub/react-slingshot
> npm run clean-dist && npm run build:html && npm run build:sass


> react-slingshot@1.0.1 clean-dist /Users/nickytonline/GitHub/react-slingshot
> npm run remove-dist && npm run create-dist


> react-slingshot@1.0.1 remove-dist /Users/nickytonline/GitHub/react-slingshot
> rimraf ./dist


> react-slingshot@1.0.1 create-dist /Users/nickytonline/GitHub/react-slingshot
> mkdir dist && cd dist && mkdir js && mkdir styles


> react-slingshot@1.0.1 build:html /Users/nickytonline/GitHub/react-slingshot
> node tools/buildHtml.js

To track JavaScript errors, sign up for a free trial at TrackJS.com and enter your token in /tools/build.html on line 10.
index.html written to /dist

> react-slingshot@1.0.1 build:sass /Users/nickytonline/GitHub/react-slingshot
> node-sass --quiet --output dist/styles src/styles/styles.scss --source-map true --output-style compressed


> react-slingshot@1.0.1 build /Users/nickytonline/GitHub/react-slingshot
> npm run build:verbose -- -s


> react-slingshot@1.0.1 build:verbose /Users/nickytonline/GitHub/react-slingshot
> node tools/build.js "-s"

Your app has been compiled in production mode and written to /dist. It's ready to roll!
unicorn:react-slingshot nickytonline$
```

Maybe I'm missing something.

coryhouse:
Correct, it doesn't since test are currently run by the build step. This was actually just a reminder to me that I'd like to start running tests upon build, but to do so, I want to disable watch for that scenario.

coryhouse:
I've updated the title accordingly.

PR #13, opened 2016-01-07
Fixes issue #1. Now the build will run tests without watching spec files.

nickytonline:
@coryhouse, with this PR you can run tests before a build (no watch) and when running npm run start they will watch as before.


coryhouse:
Thanks for the PR!

Issue #23, opened 2016-01-18
Suggestion: Update non-pure components to pure components


nickytonline:
@coryhouse There are two components in your sample app that are not pure components that could be. I'll create a PR and let me know what you think.

nickytonline:
@coryhouse, if you don't like the [route I took to build the functions](https://github.com/nickytonline/react-slingshot/commit/e409c6180506b2480050467ec54903c0d217c590#diff-0d94ee5cbbd68ec796ce79a1f8a331f7R24), I can put them inline in the pure component as they will have access to ```props```, e.g.
```
const FuelSavingsCalculatorForm = (props) => {
...
	function fuelSavingsKeypress(name, value) {
		props.actions.calculateFuelSavings(props, name, value);
	}
...
}
```

coryhouse:
Already merged so we're all set. Thanks again!

PR #24, opened 2016-01-18
Update non-pure components to pure components

nickytonline:
@coryhouse this resolves issue #23.

### DIFF
--- a/src/components/FuelSavingsApp.js
+++ b/src/components/FuelSavingsApp.js
@@ -2,79 +2,76 @@ import React, {PropTypes} from 'react';
 import FuelSavingsResults from './FuelSavingsResults';
 import FuelSavingsTextInput from './FuelSavingsTextInput';
 
-class FuelSavingsCalculatorForm extends React.Component {
-  constructor(props) {
-    super(props);
-
-    //why all the binds below? See here: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md
-    this.fuelSavingsKeypress = this.fuelSavingsKeypress.bind(this);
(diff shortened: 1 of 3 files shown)


coryhouse:
Thanks for the PR Nick! I'm on holiday today but will give this a look tomorrow!

coryhouse:
@nickytonline Okay, you've taught me something. :+1:  I've never seen this pattern of using builder functions so that you can create a stateless functional component. Do you have a link to share where you saw this documented/used?

nickytonline:
@coryhouse it's not React related just JavaScript in general. Functions can return anything including functions, i.e. higher-order functions. I just happened to call them buildFunctionX(props). I opted for this way instead of putting them inline in the pure component as I thought it was more readable but either way can be used.

```
function buildFuelSavingsKeypress(props) {
  return function fuelSavingsKeypress(name, value) {
    props.actions.calculateFuelSavings(props, name, value);
  };
}

…

const FuelSavingsCalculatorForm = (props) => {
  const fuelSavingsKeypress = buildFuelSavingsKeypress(props);
  …
}	
```

or 

```
const FuelSavingsCalculatorForm = (props) => {
  function fuelSavingsKeypress(name, value) {
    props.actions.calculateFuelSavings(props, name, value);
  }

  ...
}	
```

coryhouse:
Right, that's what threw me. I've never seen the builderX pattern. I prefer nesting the function declarations - less boilerplate since you don't have to create the wrapper and call the wrapper. I'm refactoring your PR in that direction. Hope you don't mind. You had a great point - these might as well be stateless functional components!

nickytonline:
@coryhouse no worries... cool beans. I just thought it made sense to have them all as stateless components. Thanks again for the great project. :star:

Issue #21, opened 2016-01-15
Adding the Redux dev tools?


barrystaes:
I was wondering if there are plans to add [redux-devtools](https://github.com/gaearon/redux-devtools) because it looks useful. Or should i be fine without? Not sure if its bloat or not..

I'm no expert on redux, just giving it a spin.

coryhouse:
I'm considering it but haven't had a chance to try them myself yet. If someone submits a PR I'd likely merge it.

barrystaes:
I'm nowhere yet, but since you're interested: I followed their tutorial and made a (extremely minimalistic) start at https://github.com/barrystaes/react-slingshot/tree/feature-devtools

My next step would be "Exclude DevTools from Production Builds" but i'm not sure how to do that in react-slingshot without seperating the `webpack.config.js` and `configureStore.js` files into prod/dev versions. I'd be happy to hear your take on this.

cuzzo:
@barrystaes,

Off the top of my head, you could look at webpack.hotModuleReplacementPlugin(). Then in `src/store/configureStore.js`, if module.reduxDevtools (or however you specify it) is defined, you could import the Devtool container, and then compose the reducers with Devtools.instrument(). Otherwise, do nothing.

Similarly, in  `src/containers/App.js`, you could check if module.reduxDevtools is set before adding it to the App.

Does that make sense?

coryhouse:
@barrystaes The webpack.config.js in this project is actually a factory for a webpack config. You pass it an environment and it returns a config. So no separate file is necessary. 

Anyway, thanks Barry. I want to spend some time getting comfortable with the Redux devtools. Looks like you've got a great start.

barrystaes:
Ok thanks! With your suggestions i completed my [feature-devtools](https://github.com/barrystaes/react-slingshot/tree/feature-devtools) branch. I kept the code minimal. 

Choices i made;
- A popup window instead of `redux-devtools-dock-monitor`. Less code, and stays out of the way of your UI.
- No `persistState()` added. Its usefull to reproduce bugs with `?debug_session=reproducing_weird_bug` but [thats easily added](https://github.com/gaearon/redux-devtools/blob/f53f660df2dd5af5e11aa40de176947cc187d140/README.md#use-devtoolsinstrument-store-enhancer).
- A prod/dev version for `configureStore.js` so that the debug module/component code is not included in production builds. Same goes for the window code in `src\index.js`.

[...]

PR #27, opened 2016-01-20
Add redux-devtools

barrystaes:
Adds `redux-devtools` 3.01 and `redux-devtools-log-monitor` 1.0.2 in a window.
Seperate configureStore.dev.js to avoid bundling debug code in production builds.

As discussed in https://github.com/coryhouse/react-slingshot/issues/21

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-slingshot",
-  "version": "1.0.1",
+  "version": "1.0.3",
   "description": "Starter kit for creating apps with React and Redux",
   "scripts": {
     "prestart": "npm run remove-dist",
@@ -20,6 +20,9 @@
     "test:watch": "npm run test -- --watch"
   },
   "author": "Cory House",
+  "contributors": [
+    "Barry Staes (http://barrystaes.nl/)"
+  ],
   "license": "MIT",
   "dependencies": {
     "object-assign": "4.0.1",
@@ -50,6 +53,8 @@
     "react-transform-catch-errors": "1.0.0",
     "react-transform-hmr": "1.0.1",
     "redbox-react": "1.2.0",
+    "redux-devtools": "3.0.1",
+    "redux-devtools-log-monitor": "1.0.2",
     "rimraf": "2.5.0",
     "sass-loader": "3.1.2",
     "style-loader": "0.12.3",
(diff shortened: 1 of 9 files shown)


coryhouse:
Hi Barry - I'm getting this: undefined is not an object (evaluating 'popup.location');

It's bombing on line 8 of showDevTools.

I really appreciate you creating the PR and I'm eager to try this. That said, I want to be upfront. After looking at the plumbing that's necessary for redux-devtools, I'm hesitant to incorporate this. I want to avoid confusing people by adding more complexity, and I'm not sure this will be used by the majority (though I do admit it demos really well).

Does that make sense? Am I nuts?

nickytonline:
@coryhouse , do you have a popup blocker running in your browser? I think that may be the issue.

coryhouse:
Ah, bingo. Safari was quietly surpessing it, but Chrome at least notified me. Working great in Chrome!

barrystaes:
@coryhouse actually i agree: its a lot of plumbing just to keep the debug code out of production. Apparently there is no other way to allow this code to be optimized out during build. I leave the verdict on complexity with you. That said, i think the success of react-slingshot lies in being complete but not overly complex.

barrystaes:
Perhaps its best to keep the complex in a seperate fork. Its kind of what i'm doing with my trebuchet fork..

coryhouse:
I'm going to go ahead and close this, but I'm adding a note to the FAQ that links to this thread for anyone interested in a branch that makes this happen. Thanks for your understanding and insight!

thadk:
https://github.com/barrystaes/react-trebuchet/blob/79c55de28d8e0030a7aa8f83a45be806e43f58dc/src/store/configureStore.dev.js#L18

This is actually a better, recommended way to support the devToolsExtension along with middleware (https://github.com/zalmoxisus/redux-devtools-extension) that was added in #49

coryhouse:
@thadk Seriously bizarre timing - I was literally about to commit this, but [my commit](https://github.com/coryhouse/react-slingshot/commit/e0956a02783b0615228f5e2c73252cb563fce437) is structured slightly differently. Thoughts?

thadk:
Nice! Looks like mostly the `const` is different and I'm new enough to ES6 that I don't know for sure whether replaceReducer should definitely have a `let` instead, but it seems to work either way.

PR #28, opened 2016-01-20
Fix error parameter reference in build

Zone-Ghost:


coryhouse:
Good catch. Thanks for the PR!

PR #31, opened 2016-01-21
react-slingshot is a starter kit as per project description

bettiolo:
The README.md file is lacking consistency with the project description. I would argue that this is not a framework but a set of best practices.


coryhouse:
Point taken. Thanks!

Issue #36, opened 2016-01-28
Upgrade build tools code to use ES6


coryhouse:

nickytonline:
@coryhouse, I'll get on this tonight or tomorrow and get you a PR.

coryhouse:
Awesome! Thanks so much!

PR #39, opened 2016-02-03
Upgraded the build tools to ES6. #36

nickytonline:
@coryhouse, here's the tools updated to ES6. I didn't realize that I couldn't use import in node even in v. 4.x, so I used babel-node instead for the tools.


nickytonline:
@coryhouse, I also, had to upgrade to es-lint@4.1.7 to an error I was getting when linting, [TypeError: Cannot read property 'visitClass' of undefined](https://github.com/babel/babel-eslint/issues/243).

coryhouse:
This is perfect! Thanks so much!

PR #47, opened 2016-02-06
connect-history-api-fallback

dotnetnorm:
Adds connect-history-api-fallback to the project, which will allow SPAs
to always return the index.html file for any url entry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "chai": "3.4.1",
     "cheerio": "0.19.0",
     "colors": "1.1.2",
+    "connect-history-api-fallback": "1.1.0",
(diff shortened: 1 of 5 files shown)


coryhouse:
Thanks! I'll give this a look on Monday.

vasyabigi:
Thank you! Works well :+1:

rafaelquintanilha:
Thanks for this. The only drawback is that when building for production the assets have a wrong path.

coryhouse:
This was added as part of 2.0 release via the [PR to add React Router](https://github.com/coryhouse/react-slingshot/pull/61).  Thanks for the PR though!

PR #51, opened 2016-02-08
[DOC] npm run open, build:sass commands not present

bf4:


coryhouse:
Good catch. Thanks!

bf4:
Thanks. Yeah, I now see there's a few more.

Issue #48, opened 2016-02-06
Including third party modules


wormyy:
Hello I am starting an app with your starter kit. I wanted to include normalize.css and react-toolbox but I am getting these errors during the build:

```
ERROR in ./~/normalize.css/normalize.css
Module parse failed: /home/jules/projects/tourbnb-frontend/node_modules/normalize.css/normalize.css Line 9: Unexpected token {
You may need an appropriate loader to handle this file type.
|  */
| 
| html {
|   font-family: sans-serif; /* 1 */
|   -ms-text-size-adjust: 100%; /* 2 */
 @ ./src/index.js 29:0-24

ERROR in ./~/react-toolbox/lib/app/App.js
Module not found: Error: Cannot resolve 'file' or 'directory' ./style in /home/jules/projects/tourbnb-frontend/node_modules/react-toolbox/lib/app
 @ ./~/react-toolbox/lib/app/App.js 11:13-31
```

This is my index.js: 
```
import React from 'react';
import {render} from 'react-dom';
import { Provider } from 'react-redux';
import ToolboxApp from 'react-toolbox/lib/app';
import App from './containers/App';
import configureStore from './store/configureStore';
import './styles/styles.scss'; //Yep, that's right. You can import SASS/CSS files too! Webpack will run the associated loader and plug this into the page.
import 'normalize.css';

const store = configureStore();

render(
  <Provider store={store}>
    <ToolboxApp>
      <App />
    </ToolboxApp>
  </Provider>, document.getElementById('app')
);
```

I am fairly new to react and webpack, so I am sorry if it's something really obvious.

coryhouse:
I'm not familiar with normalize.css, but it appears the issue is either:
1. It's referencing a file that it cannot find. Check your paths.
2. It's referencing a file type that webpack isn't currently setup to load. 

If it's 2, then check you'll just need to find the appropriate webpack loader for the file being loaded.

wormyy:
But normalize is just a css file, that should be handled by this loader, shouldn't it?
```
loaders.push({
      test: /(\.css|\.scss)$/,
      include: path.join(__dirname, 'src'),
      loaders: ['style', 'css?sourceMap', 'sass?sourceMap']
    });
```

I realized that if I change path `path.join(__dirname, 'src')` to `path.join(__dirname)` it loads properly. 

However the react-toolbox problem still persists.

coryhouse:
It's looking in /src so you'll need to place your files there too.

[...]

PR #55, opened 2016-02-20
Make slingshot compatible with the React Toolbox (http://react-toolbox.com/)

egbertp:
* Tell webpack to resolve .jsx .scss .js and .json files automatically.
* Tell the CSS loader to include modules (@import ...)

Now this slingshot is compatible with the React Toolbox (http://react-toolbox.com/)

This fixes coryhouse/react-slingshot#48

Error message I saw before the improvement:
```
Error: Cannot resolve 'file' or 'directory' ./style in /<YOUR_PROJECT_FOLDER>/node_modules/react-toolbox/lib/button
```

If I list the contents of this folder, I see a file `style.scss`, but not a file `style` (see bellow):
```
$ ls -l ~/<YOUR_PROJECT_FOLDER>/node_modules/react-toolbox/lib/button 
-rw-r--r--  1 egbert  staff  6497 Jan 26 14:10 Button.js
-rw-r--r--  1 egbert  staff  5594 Jan 26 14:10 IconButton.js
drwxr-xr-x  3 egbert  staff   102 Jan 26 14:10 __test__
-rw-r--r--  1 egbert  staff  1075 Jan 26 14:10 _config.scss
-rw-r--r--  1 egbert  staff   528 Jan 26 14:10 index.js
-rw-r--r--  1 egbert  staff  3628 Jan 26 14:10 style.scss
```

Now webpack automatically resolves `style` to `style.scss`.

If there is anything unclear: please feel free to ask!


coryhouse:
Thanks for the PR! I want to assure I understand the new loader config:

`loaders.push({test: /(\.css|\.scss)$/, loaders: ['style', 'css?sourceMap&modules&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]', 'sass?sourceMap']});`

You added this:

`&modules&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]`

Oof. That line is cryptic. After reading the docs, the line you added tells Webpack to do the following:

1. Load any @import statements and give them a name that begins with the name assigned
2. Assign the styles to the local selector scope (rather than global)
3. Export the local selectors with add a hash on the end of the selector
4. Generate a sourcemap for debugging

Is that a valid summary? 

One blocking issue - the production loader config needs updated as well.

Finally, is there a reason you chose React Toolbox over http://www.material-ui.com/? (I'm currently using the latter on a separate project).

Thanks!

FlynnP:
I'm trying to get react-toolbox working with a more recent version of react-toolkit and I'm having trouble. It looks like I'm getting the same error but the webpack config has changed so this fix doesn't seem to work. Any ideas?

Also with regards to the question by @coryhouse of material-ui vs react-toolbox, I tried using material-ui but I was quickly stymied by it's lack of support for css class customization. Since all of their styling is inline (which I can see some value to) and doesn't expose all of the nodes one might want to modify, I ended up having to create an insane mix of js and css overrides that were a pain to keep track of.

kwelch:
@FlynnP To echo your point, one of the biggest advantages of react-toolbox is that they use CSSModules and decorators to allow you to either take the component as styled or the un-styled copy. I went through a lot of research today and found that they were my preferred approach over a lot of the UI frameworks out there. Toolbox is extremely similar to material-ui, in my opinion, in respects to the styling you get and theming, but the edge to me feel on toolbox when I say that their dev approach was much closer to what I would expect if I were writing it. 

One other thing I did to make that cryptic loader query params more approachable was using JSON.stringify and using object notation.

coryhouse:
I'm open to a PR on this if there's a way to add support that's less complicated than the initial request on this thread. @FlynnP I can't tell at a glance what you've changed due to formatting, but would be open to reviewing a separate pull request if you're willing. Thanks!

kwelch:
I recommend closing, as this is out of scope for Slingshot at this time. There are quite a few CSS options at this point. 

@coryhouse, Do you agree?

coryhouse:
Agreed. Sorry I let this languish. Closing since the code change is too cryptic and edge-case to justify.

coryhouse:
@retorquere I'm not longer using material-ui. Sorry, I don't have any reference handy but don't recall it being a painful set up.

PR #72, opened 2016-03-10
webpack-hot-middleware/client reload params configured to true

Boychenko:
I think it will be better DX to have full reload when webpack gets stuck.[config](https://github.com/glenjamin/webpack-hot-middleware#config)
This very useful for case when HMR fails due to [stateless functional components](https://github.com/gaearon/babel-plugin-react-transform/issues/57#issuecomment-167675819). It's easy to reproduce for top level components like `App` or `AboutPage`

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -40,7 +40,7 @@ const getEntry = function (env) {
   const entry = [];
 
   if (env === developmentEnvironment ) { // only want hot reloading when in dev.
-    entry.push('webpack-hot-middleware/client');
+    entry.push('webpack-hot-middleware/client?reload=true');
   }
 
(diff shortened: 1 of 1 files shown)


coryhouse:
Great call! Didn't even know this option existed. My only hesitation is some people will be confused by how sometimes a full reload occurs. I believe the rule is top-level components will always require a reload. Are there other instances this would occur? Regardless, I agree this is a win, so merging. Thanks for the PR!

Boychenko:
I didn't find description about reload conditions. Documentation doesn't say something concrete. It says that reload happens when webpack stucks. Also this option was mention in breaking changes on version 2.0.0.

Issue #71, opened 2016-03-10
Integrate CSS-Modules


bresson:
Would be great if Slingshot could integrate CSS-Modules:
https://github.com/css-modules/webpack-demo

coryhouse:
I'm open to it if someone submits a pull request.

kwelch:
As it has been said in #77, there are many options and at this point there is no particular stance begin taken in react-slingshot.

PR #77, opened 2016-03-20
adding react-css-modules support

atfzl:
Adds enhancement from https://github.com/coryhouse/react-slingshot/issues/71.

But tests are failing. Babel is trying to parse the .scss and .css files in src.
I think that babel was always parsing all the files in src folder and not just *.spec.js files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "connect-history-api-fallback": "1.1.0",
     "object-assign": "4.0.1",
     "react": "0.14.7",
+    "react-css-modules": "3.7.6",
     "react-dom": "0.14.7",
     "react-redux": "4.4.0",
     "react-router": "2.0.0",
--- a/src/components/FuelSavingsResults.js
+++ b/src/components/FuelSavingsResults.js
@@ -1,6 +1,9 @@
 import React, {PropTypes} from 'react';
 import NumberFormatter from '../businessLogic/numberFormatter';
 
+import CSSModules from 'react-css-modules';
+import styles from './FuelSavingsResults.scss';
+
 //This is a stateless functional component. (Also known as pure or dumb component)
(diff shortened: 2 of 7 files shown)


coryhouse:
Thanks @atfzl! I'll give this a look shortly.

jeremyadavis:
@atfzl The tests failing should probably be fixed with the ignore-styles PR I opened yesterday.

atfzl:
@jeremyadavis Thanks for pointing this out. The problem was gone after adding ignore-styles.

@coryhouse but there is another problem in [FuelSavingsResults.spec.js](https://github.com/coryhouse/react-slingshot/blob/master/src/components/FuelSavingsResults.spec.js#L36).
It is trying to find the html element with css class .fuel-savings-label but the name of the css class is changed by react-css-modules. So maybe we should remove these kinds of tests.

coryhouse:
I'm open to suggestions on how to do that better. I'm merely wanting to display an example of asserting specific resulting markup.

liady:
@coryhouse @atfzl Add this code in `testSetup.js`:

```js
// css modules hook
var hook = require('css-modules-require-hook');
var sass = require('node-sass');
hook({
  extensions: [ '.css', '.scss' ],
  preprocessCss: (data, file) => sass.renderSync({ file }).css,
  generateScopedName: '[local]'
});
```
This will guarantee that the class names will not be changed during tests.

coryhouse:
After thinking about this more, there are over a dozen different ways to handle CSS with React, so I don't think having an opinion on how Slingshot handles this makes sense at this time.

Maxim-Filimonov:
It's been almost a year now since this PR. It's possible to integrate using just babel - https://github.com/gajus/babel-plugin-react-css-modules
Is it feasible to reconsider integration of CSS modules?

PR #78, opened 2016-03-20
Only log success after writing finishes.

firstdoit:
Small improvement, but semantically important for people learning :)


coryhouse:
@firstdoit Great catch! Thanks!

Issue #65, opened 2016-03-03
Nested routes are not working when URL is refreshed


CreativeManix:
```
<Route path="/" component={App}>
    <IndexRoute component={FuelSavingsPage} />
    <Route path="about" component={AboutPage}/>
    <Route path="/auth"  component={AuthLayout} >
		<Route path="login" component={LoginPage}/>
		<Route path="atkn" component={AcceptTokenPage}/>
    </Route>
    <Route path="*" component={NotFoundPage} />
  </Route>
```

I have above configuration, when I navigate to /auth/login it works. however when I refresh the page, history fallback lib is not rendering the component instead it shows blank screen!

Should I do anything specifically?

atfzl:
try this
```
<Route path="/" component={App}>
    <IndexRoute component={FuelSavingsPage} />
    <Route path="about" component={AboutPage}/>
    <Route path="auth" component={AuthLayout} >
        <IndexRoute component={LoginPage}/>
        <Route path="atkn" component={AcceptTokenPage}/>
    </Route>
    <Route path="*" component={NotFoundPage} />
  </Route>
```

CreativeManix:
Sorry, It is not working. If I navigate using <Link> I am able to reach  _/auth/atkn_ however when I refresh my browser it is showing blank screen.

atfzl:
Can you give the url to your repo.

coryhouse:
@CreativeManix Get latest and try again. I added a leading slash in index.html for bundle.js. I believe this will fix your issue.

coryhouse:
@CreativeManix - Is this issue resolved when you get latest?

coryhouse:
Closing since this should have been fixed with a recent commit.

jforaker:
This happens in the latest version as of 6/30/2016, and is undesirable...  any suggestions?  

The pull request above references a file that is no longer in the repo.

coryhouse:
Great catch @jforaker. Bummer. This is my miss. The new structure for generating index.html via html-webpack-plugin caused this. I'm looking into a fix, but in the meantime, 4.0.0 will continue to work. I haven't cut a release in awhile so it's quite a few commits behind master.

thatkit:
@CreativeManix I want to follow up on @coryhouse's answer on the leading slash solution.

In order to fix the issue by adding a leading slash to `<script src="index.js"></script>` (the react script imported in `index.html`).

`webpack.config.js`
```javascript
const config = {
[...]

PR #79, opened 2016-03-20
Make bundle URL relative to index.html

firstdoit:
This allows a deploy to non-root URLs to be seamless.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,6 +6,6 @@
   </head>
   <body>
     <div id="app"></div>
-    <script src="/bundle.js"></script>
+    <script src="bundle.js"></script>
   </body>
 </html>


coryhouse:
I added the leading slash to [resolve this](https://github.com/coryhouse/react-slingshot/issues/65). That said, he never shared if that helped. If we can confirm that nested routes can reload successfully during dev without the leading slash then I can accept this PR.

Thanks!

rndstr:
I think `bundle.js` can be either absolute or relative but it must be the same as `styles.css` in tools/buildHtml.js which it is currently not. Otherwise it can break in production.

For me it needs to be absolute since I have entry points that are on the second level (app at example.com and client can come in at example.com/en/something). And even if you have your app at a sub path like example.com/calculator/, if you want to support clients loading your app from example.com/calculator/en/something you will need to make it absolute with the current setup.

I opened a PR to fix the current state (`styles.css` being not the same as `bundle.js`) but feel free to close if you would rather have both relative.

coryhouse:
@firstdoit Can I close this? Does #143 resolve this for you?

firstdoit:
Not sure if everything absolute is what most people expect (or even how it solves the problem of deploying to multi-level paths, given that now we're restricted to root!), but sure, any decision is better than no decision!

rndstr:
@firstdoit I see what you mean

Relative solves the problem of deploying to multi-level paths. Absolute solves the problem of reloading an app after navigating to a multi-level path.

But, I think react-slingshot does at the current state not support reloading the site after navigating, correct me if I'm wrong. Which would make my point moot for having it absolute. If you do have navigation&reload support (which my application has) then I think it is inevitable to have it absolute.

firstdoit:
AFAIK the expected behaviour in a React (or rather any SPA) app is to have the app itself handle all navigation and routing. I agree that for anything non-trivial, specially using `History`, you'll eventually have to bolt down the paths. In this case, however, I'd either document this detail or make it simple to add a `base path`

PR #82, opened 2016-03-21
Added imported styles and test case for AboutPage component

jeremyadavis:
Adds arbitrary change to AboutPage component to demonstrate how to import a stylesheet and adds an accompanying unit test to validate that the imported stylesheet isn't picked up by mocha.


coryhouse:
Thanks so much for the PR Jeremy!

PR #85, opened 2016-03-29
Fix compose error in configure store dev version

shallker:
fix error  'compose' is not defined in configureStore.dev.js


coryhouse:
Thanks!

PR #86, opened 2016-03-30
Now there is a pre-check to see if there are any tests to run.

### LINKED ISSUE #40: cannot resolve path (or pattern) './src/**/*.spec.js' #21
How can fix this? please help..

nickytonline:
@coryhouse, this fixes #40 . The only thing I don't like is that I have to specify the glob in the `pretest` and `test` scripts. I can set the process.env.testsGlob to the pattern passed in from the runTestsPreFlight.js script, but when the `test` script runs, I do not know how to pass it to mocha. Open to suggestions or we can go as is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "prebuild": "npm run clean-dist",
     "build": "npm-run-all --parallel test build:html build:js",
     "postbuild": "npm run open:dist",
+    "pretest": "babel-node tools/runTestsPreFlight.js --glob \"./src/**/*.spec.js\"",
     "test": "cross-env NODE_ENV=test mocha --reporter progress --compilers js:babel-core/register --recursive \"./src/**/*.spec.js\" --require ignore-styles",
     "test:watch": "npm run test -- --watch"
   },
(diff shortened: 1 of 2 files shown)


coryhouse:
Nice work! This is just what I was picturing, but now that I see the impact, I'm concerned about the extra complexity we're taking on. Due to my initial decision to place .spec files alongside files under test, we have the following added complexity:

1. --recursive flag on test script
2. glob on test script
3. preflight task which calls separate script and duplicates glob
4. glob package dependency
5. preflight script that does the check
6. Challenges with wiring up code coverage in the future (similar struggles)

Now that I see the full impact, I'm considering alternatives:

1. Should we [run Mocha programmatically](https://github.com/mochajs/mocha/wiki/Using-mocha-programmatically) instead?
2. Should we simply move tests to the default Mocha folder (test) to simplify? How much do people really care where tests are placed?

nickytonline:
TLDR, I'm OK with leaving the tests as is.

Honestly, I thought it was a weird request to begin with. I think mocha exploding in your face is OK.
1. A project should have tests.
2. Command line is simpler. Doing mocha via code in my opinion will complexify things. 
3. I think code coverage will be simpler this way via iSparta,  remap-Istanbul or a similar tool.

coryhouse:
Thanks for understanding! I'm going to go ahead and close this and the issue for now. That said, you did this exactly the way I envisioned and I really appreciate you taking the time. :+1:  If it becomes possible to message to the user without taking on extra complexity in the future we can always add it.

nickytonline:
I've moved this to https://github.com/nickytonline/react-slingshot/tree/bug-check-for-tests if anyone wants to use it.

PR #89, opened 2016-03-31
Remove babel-eslint and disable stage-1 support

kwelch:


coryhouse:
Hi @kwelch. Thanks for the PR! I'm going to hold off on this since I'm trying to gauge people's opinion on removing stage-1 support. I'm hesitant to take something away that people are using. The only thing babel-eslint (and stage-1 support) are costing us in 1 extra dependency, so I want to gauge whether people think it's worth moving.

coryhouse:
I just committed [a similar change](https://github.com/coryhouse/react-slingshot/commit/94e49bcb45f2539191d8d58a2af8c1a3103a1d2c) that included migrating to eslint-watch for enhanced command line output. That said, thanks for the PR Kyle!

Issue #60, opened 2016-02-24
Add proxy for API calls


prpatel:
In the past, I've used webpack-dev-server proxy to pass API calls from my dev React app to a running API server (Rails or Node, but that does not matter TBH). 
I tried adding the webpack-dev-server proxy to srcServer.js:
proxy: {
          '/api/*': {
            target: 'http://localhost:3000/',
            secure: false
           // ...
then I realized that this project wraps webpack-dev-middleware around browserSync, so it does not work. Adding the browserSync proxy does not serve the web/React assets properly - all this stuff should live in the react-slingshot project, and _only_ API calls should be proxy'd to the API server. Suggestions for making this work?

coryhouse:
Great question. Can you point me to a working configuration example?

prpatel:
@coryhouse I assume you're asking for an example in webpack-dev-server. Have a look at this file in one of my repos:
https://github.com/prpatel/react-workshop-day1/blob/master/Lab12/webpack.config.js

coryhouse:
So to be clear, this is an alternative to enabling CORS on local APIs, right? I suspect it's possible to pull this off along with browser-sync, but I haven't had time to hunt for an example yet.

prpatel:
Actually, it is more about a seamless development experience / setup than CORS. Allow me to elaborate. I'm running my Rails/Node/Java/Python/etc backend server and want to create a SPA using React. During development, I need to either do one of these things to make this work:
1) use ENV vars to set the base URL to my backend API server
2) use magic vars (let authUrl = "http://localhost:3000";) that are managed using (1) or some other method
3) Proxy API calls with some simple configuration through browsersync+webpack in this project's setup

IMO, (3) is not the easest, but is the most simple. From the research I've done into this, I think the answer may be to invert the browsersync+webpack setup:
https://www.npmjs.com/package/browser-sync-webpack-plugin
but I'm not a browersync expert, and haven't had time to dig into its guts to determine how this is possible - was hoping that someone here had already tackled this and had a quick answer.

vssathish:
I am doing proxy using proxy-middleware. Modified srcServer.js


// This file configures the development web server
// which supports hot reloading and synchronized testing.

// Require Browsersync along with webpack and middleware for it
import browserSync from 'browser-sync';
import webpack from 'webpack';
import webpackDevMiddleware from 'webpack-dev-middleware';
import webpackHotMiddleware from 'webpack-hot-middleware';
_**import url from 'url';
[...]

PR #93, opened 2016-04-03
Proxy for Webpack.

vssathish:
pull request for https://github.com/coryhouse/react-slingshot/issues/60#event-590595742

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,10 @@
     "webpack": "1.12.11",
     "webpack-dev-middleware": "1.4.0",
     "webpack-hot-middleware": "2.6.0",
-    "yargs": "3.32.0"
+    "yargs": "3.32.0",
+    "url": "0.11.0",
+    "http-proxy-middleware": "0.11.0",
+    "proxy-middleware": "0.15.0"
   },
   "keywords:": [
     "react",
--- a/tools/srcServer.js
+++ b/tools/srcServer.js
@@ -11,8 +11,15 @@ import webpackDevMiddleware from 'webpack-dev-middleware';
 import webpackHotMiddleware from 'webpack-hot-middleware';
 import webpackConfigBuilder from '../webpack.config';
 
+import url from 'url';
+import proxy from 'proxy-middleware';
+
 const webpackConfig = webpackConfigBuilder('development');
 const bundler = webpack(webpackConfig);
+
+// Proxy Poptions
+var proxyOptions = url.parse('http://evcadmin.test.netflix.net/REST');
+    proxyOptions.route = '/REST';
 
 // Run Browsersync and use middleware for Hot Module Replacement
 browserSync({
@@ -36,6 +43,8 @@ browserSync({
 
       // bundler should be the same as above
       webpackHotMiddleware(bundler),
+
+      proxy(proxyOptions),
 
       historyApiFallback()
     ]
--- a/tools/srcServer.js
+++ b/tools/srcServer.js
@@ -17,8 +17,8 @@ import proxy from 'proxy-middleware';
 const webpackConfig = webpackConfigBuilder('development');
 const bundler = webpack(webpackConfig);
 
-// Proxy Poptions
-var proxyOptions = url.parse('http://evcadmin.test.netflix.net/REST');
+// Proxy options. Update the proxy url.
+var proxyOptions = url.parse('http://proxyserver.net/REST');
     proxyOptions.route = '/REST';
 
 // Run Browsersync and use middleware for Hot Module Replacement


coryhouse:
Thanks for the PR! This is quite clean and straightforward. Only issue I see off-hand is the packages in package.json should be sorted alphabetically.  I need some time to try this myself to understand the implications so please bear with me.

raythree:
This is exactly what I was trying to do. I'm migrating from gulp/browserify to slingshot-based/webpack builds. All of my SPAs talk to backend micro-services, so with gulp/browserify I was generating the output directly into the server/public...so I could do "gulp watch"...edit code, refresh... it was a little slow, but it was managable... the proxy-middleware looks like the perfect solution here.

kwelch:
This solution works well, and I use it in my solutions as well as a proxy file that includes environments and all setup. Is this something that most/all people will need though? Does it make sense to put in more docs on when and why to use it?

raythree:
As most SPAs talk to backend services I think it's very useful, but very site/app specific. I don't think anything is needed from slingshot, except maybe document how to configure it. BTW, I found that http-proxy-middleware was very convenient, especially when there are multiple routes that need to get proxied, and no need to import 'url'. In tools/srcServer:

```javascript
import proxy from 'http-proxy-middleware';

const serverProxy = proxy(['/login', '/api'], {
    target: 'http://localhost:8000',
    changeOrigin: false, // set true for hosted sites
    logLevel: 'debug'
});
// now insert serverProxy into browsersync config into server: { middleware:[.., proxyServer, ...] 
```

coryhouse:
I'd welcome docs on how to use it. Since each person's configuration is going to be unique, it seems docs make more sense than a commit.

conoverm:
thanks to @raythree I got a proxy up. I'm not new to browsersync, but very new to webpack, react, et al. I had no idea where to put the proxy config because I was unsure what was actually serving the app.

Incidentally, http-proxy-middleware is now part of wepback (which served to confuse me a bit further).
https://github.com/webpack/webpack-dev-server/blob/master/package.json#L13

kwelch:
Is this something we want to bundle in with slingshot or is this better suited as a documentation update?

coryhouse:
I lean toward docs since the config will differ by project. A link to this PR and @raythree excellent comment would be a great start.

kwelch:
I currently don't have the availability to write up the documentation for this, but one thing I would recommend, for anyone that is working on features similar to this, to put this type of configuration into a separate file. This will centralize this node server middleware/plugin configuration for both the dev and dist server and ensure consistency. 

I have done this with some of my projects as simply as:
```
import proxy from 'http-proxy-middleware';

module.exports = function (app) {
  app.use(proxy(['/login', '/api'], {
    target: 'http://localhost:8000',
    changeOrigin: false, // set true for hosted sites
    logLevel: 'debug'
  }));
};
```

mikedevita:
Just commenting on my approach on how I handle this along with a .NET backend that uses NTLM and requires the use of agentkeepalive. You could even do like @kwelch suggested and pull the proxy logic out and include it as a middleware all the time it just doesn't do anything if you don't have it enabled. I'm not sure if there is a better way to pull the configs into `package.json` other than from the direct import into `srcServer.js`. Speaking of which has anyone given the thought of pulling these configs out into the base dir either in a `config.js` file or into `package.json`?

I do know that the order of the proxyMiddleware is important, i always put it at the beginning of the browserSync middleware array.

it'd be great if there was a yeoman generator for this slingshot boilerplate that asks you a few questions, like do you need a proxy, whats the endpoints, etc and builds it out for you.

Anyways this is my steps as I have documented:

## install required proxy packages

```shell
yarn add -D http-proxy-middleware agentkeepalive
# -- or --
npm install --save-dev http-proxy-middleware agentkeepalive
```

## In package.json add the following...

```json
   "proxy": {
    "enabled": true,
    "routes": "/api",
    "options": {
      "target": "http://localhost:6741",
      "changeOrigin": false,
      "logLevel": "debug"
    },
    "keepAlive": {
      "enabled": true,
      "options": {
        "maxSockets": "100",
        "keepAlive": true,
        "maxFreeSockets": "10",
        "keepAliveMsecs":"1000",
        "timeout": "60000",
        "keepAliveTimeout": "30000"
      }
    }
  },
```

## in tools/srcServer.js replace with the following

```js
// This file configures the development web server
// which supports hot reloading and synchronized testing.

// Require Browsersync along with webpack and middleware for it
import browserSync from 'browser-sync';
// Required for react-router browserHistory
// see https://github.com/BrowserSync/browser-sync/issues/204#issuecomment-102623643
import historyApiFallback from 'connect-history-api-fallback';
import webpack from 'webpack';
import webpackDevMiddleware from 'webpack-dev-middleware';
import webpackHotMiddleware from 'webpack-hot-middleware';
import config from '../webpack.config.dev';
import proxy from 'http-proxy-middleware';
import Agent from 'agentkeepalive';
import packageJson from '../package.json';

const bundler = webpack(config);
let middleware = [];

if (packageJson.proxy.enabled) {
  let options = packageJson.proxy.options;

  if (packageJson.proxy.keepAlive.enabled) {
    const keepaliveAgent =  new Agent(packageJson.proxy.keepAlive.options);

    const onProxyRes = function (proxyRes, req) {
      const key = 'www-authenticate';
      proxyRes.headers[key] = proxyRes.headers[key] && proxyRes.headers[key].split(',');
    };

    options = {
      ...options,
      agent: keepaliveAgent,
      onProxyRes: onProxyRes
    };
  }
  
  const proxyMiddleware = proxy(
    packageJson.proxy.routes,
    options
  );

  middleware.push(proxyMiddleware);
}

middleware = [
  ...middleware,
  historyApiFallback(),
  webpackDevMiddleware(bundler, {
    // Dev middleware can't access config, so we provide publicPath
    publicPath: config.output.publicPath,
    // These settings suppress noisy webpack output so only errors are displayed to the console.
    noInfo: true,
    quiet: false,
    stats: {
      assets: false,
      colors: true,
      version: false,
      hash: false,
      timings: false,
      chunks: false,
      chunkModules: false
    },

    // for other settings see
    // https://webpack.js.org/guides/development/#using-webpack-dev-middleware
  }),

  // bundler should be the same as above
  webpackHotMiddleware(bundler)
];

// Run Browsersync and use middleware for Hot Module Replacement
browserSync({
  port: 3000,
  open: false,
  ui: {
    port: 3001
  },
  server: {
    baseDir: 'src',
    middleware
  },

  // no need to watch '*.js' here, webpack will take care of it for us,
  // including full page reloads if HMR won't work
  files: [
    'src/*.html'
  ]
});
```

nickytonline:
This PR is quite old, but just following up. From the comments, I think the consensus was to document this instead?

kwelch:
Yes, I agree.

nickytonline:
Closing this and will add to #60 that we just document this.

PR #103, opened 2016-04-09
Implements new testing utilities and additional test cases

jeremyadavis:
Took a bit longer than I expected, but the rabbit hole of React testing went deeper than I thought.

The React team recommends using [shallow rendering](https://facebook.github.io/react/docs/test-utils.html#shallow-rendering). It allows components to be tested in isolation and doesn't require a DOM to be present.

Although React Addons have their own utility methods to do shallow rendering, I chose to use the [Enzyme](https://github.com/airbnb/enzyme) test utilities to accomplish it. Enzyme is less verbose, simpler to use and has decent documentation.

[Sinon](http://sinonjs.org/) is used to spy on functions.

I added tests for Redux Actions, Reducers and the Store. The Store tests are more integration type tests.  They are probably pointless since there are tests on Actions and Reducers now, but I think it's helpful seeing how a Redux enabled app is just a collection of action creators manipulating a single state.

The Container had to be slightly modified (along with an eslint rule) to test the unconnected component, which is what [Redux recommends](http://redux.js.org/docs/recipes/WritingTests.html)

I attempted to demonstrate various ways to go about composing test cases and assertions. Most examples used the Expect style, so that is what I predominately used, but one component follows the Should style.

I kept most of the test cases overly isolated to be easily manipulated by someone learning. In the real-world I would have employed more factory functions or use of Mocha's before hook.

Let me know if you have any questions.


coryhouse:
Ha! Wow, that's awesome! :100:  I literally have a commit ready for adding Enzyme, but you took it **_way_** farther than I did! This is seriously impressive and I'm sure took many hours. Thanks so much! :+1: 

Love seeing that you choose the same test stack as me! Only difference is I was using Expect instead of Chai (it's less verbose), but Chai is quite similar and more popular so I'm totally fine with that. 

I'm really happy with this, so I'm going to merge as is. :) But here's some things I'd like to tweak in a follow-up commit:

1. Rather than disable [this rule](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-named-as-default.md) globally, perhaps we should disable it in this one case? Another option is to [just import it with a different name in routes.js](https://github.com/benmosher/eslint-plugin-import/issues/183). But I prefer the name we're using now. Torn on this one.
2.  In some cases, [you're setting a constant that appears to be duplicated](https://github.com/coryhouse/react-slingshot/pull/103/files#diff-4bca6719738f133e2b3ea7409690df71R9). Looks worth moving to the top to centralize. Perhaps use beforeEach?
3. [This data structure](https://github.com/coryhouse/react-slingshot/pull/103/files#diff-02bd98750a1aebde7e4b137d63ba3450R88) is repeated quite a few times. Likely worth centralizing. Likely a good place to use the [Fluent Build Pattern](https://www.kenneth-truyers.net/2013/07/15/flexible-and-expressive-unit-tests-with-the-builder-pattern/)
4. Rather than [hard coding a copy of initial state](https://github.com/coryhouse/react-slingshot/pull/103/files#diff-dd51c881cd96ddb26fdf1fbb22410ff8R10), why not add a reference to it here? I see nothing wrong with leveraging it for test bootstrapping. Helps avoid maintenance overhead later...
5. Inconsistent [verb use here](https://github.com/coryhouse/react-slingshot/pull/103/files#diff-1984f216d641a70fc07b678d10bc6064R23). Sometimes says display in test title, other times gives.

jeremyadavis:
Glad you liked it and that you were heading in the same direction as well.

I think I just stuck with Chai because that's what you originally had, plus I can demo the 'should' style without additional dependencies.

1. Yeah probably best to single-line ignore it. That was more of a rush fix because I just wanted to get the PR submitted but didn't want any warnings. My personal preference these days is to have Redux containers call dumb components instead of the self call like it is now. Just keeps the purpose of the file more focused on wiring up state, actions and props and not about the UI.

 2. Yeah. I chose to be overly explicit with stuff in hopes of making it easier for people to understand and tweak. For my app, I build getWrapper() functions that I call and can send parameters to (like sinon.spys or prop states). Any changes really are a matter more of what your goals are for slingshot. I know I originally chose it to learn React/Redux because it felt more approachable than the top yeoman ones at the time. Whatever your prefer, I'm game.

3. Same answer as 2

4. Agreed

5. I probably did that because the 'displays' refer to a values text whereas the 'gives' refer to a class name being assigned. But I'd be surprised if I keep that all consistent. My goal was to at least start all test cases with 'should' which can be harder than you'd think and lead to some awkward phrasing.

PR #113, opened 2016-04-20
Added redux logger middleware

jeremyadavis:
I prefer to use this way to monitor my state and I think it would be good candidate for the starter kit since it shows how the action modified the state object.

Allows people to see Redux without getting Redux Dev tools installed.
Demonstrates exactly how to add middleware within compose so should be easier for people to add something like thunk.

Bumped the react-redux version to avoid an npm dependency warning for react 0.14

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -59,6 +59,7 @@ Unfortunately, scripts in package.json can't be commented inline because the JSO
 |react-redux|Redux library for connecting React components to Redux |
 |react-router|React library for routing |
 |redux|Library for unidirectional data flows |
+|redux-logger|Logs Redux actions and states to console |
 |babel-cli|Babel Command line interface |
 |babel-core|Babel Core for transpiling the new JavaScript to old |
 |babel-eslint|Connects Babel and ESLint so ES6 code can be linted |
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "object-assign": "4.0.1",
     "react": "15.0.1",
     "react-dom": "15.0.1",
-    "react-redux": "4.4.0",
+    "react-redux": "4.4.5",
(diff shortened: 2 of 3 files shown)


kwelch:
Since this is a teaching/starting point for react, I recommend updating docs to reflect what the new package is doing. This way even if the pull request does not get approved it can help someone integrate into their project and understand why they may want to integrate it.

jeremyadavis:
Yep, I forgot to update it. Thanks.

@coryhouse If you think something like this needs a larger documentation call out like your 'What about Redux Devtools' section or if it is potentially annoying to people and need to know how to undo it, let me know and I can update the FAQ further.

coryhouse:
Hrmm. Why would one want to use this when the Redux dev tools (which are already supported) provide this information and so much more?

jeremyadavis:
It's not going to be a huge value add, and might boil down to just my preference. 

I prefer to see this output in my console over using Redux dev tools because it shows me the previous state, the action and the mutated state from the reducer. I find it easier to inspect my state from the console than dev tools.

Obviously you don't get timetravel debugging features of dev tools, but I rarely use that and if I do I still keep dev tools running.

Also, I thought to get dev tools working you still need to install a monitor like the Chrome extension, so I figured the logger would add some value without additional dependencies.

coryhouse:
I can totally see how this can be useful, but I'm concerned about the amount of noise it creates within the console. I agree it's a useful tool, but I don't believe it should be enabled by default since I'd only use it in a debugging situation. That's why I prefer continuing to recommend the Redux dev tools instead. I'm concerned logging everything to the console could annoy as many people as it pleases. And since it's trivial to add, I think I'll leave it out for now. 

That said, thanks for the PR. :)

jeremyadavis:
fair enough

PR #115, opened 2016-04-21
Move "connect-history-api-fallback" to devDependencies

nhducit:
Move "connect-history-api-fallback" to devDependencies


coryhouse:
Great point. Thanks!

Issue #116, opened 2016-04-23
npm install failure


znowden:
After downloading react-slingshot and calling "npm install" I get this error:

    313437 info install react-slingshot@3.0.1
    313438 info postinstall react-slingshot@3.0.1
    313439 info prepublish react-slingshot@3.0.1
    313440 verbose validateInstall loading C:\Users\znowden\Desktop\react-slingshot\package.json for validation
    313441 verbose stack Error: The package react@15.0.1 does not satisfy its siblings' peerDependencies requirements!
    313441 verbose stack     at C:\Program Files\nodejs\node_modules\npm\lib\install.js:125:32
    313441 verbose stack     at C:\Program Files\nodejs\node_modules\npm\lib\install.js:268:7
    313441 verbose stack     at C:\Program Files\nodejs\node_modules\npm\node_modules\read-installed\read-installed.js:142:5
    313441 verbose stack     at C:\Program Files\nodejs\node_modules\npm\node_modules\read-installed\read-installed.js:263:14
    313441 verbose stack     at cb (C:\Program Files\nodejs\node_modules\npm\node_modules\slide\lib\async-map.js:47:24)
    313441 verbose stack     at C:\Program Files\nodejs\node_modules\npm\node_modules\read-installed\read-installed.js:263:14
    313441 verbose stack     at cb (C:\Program Files\nodejs\node_modules\npm\node_modules\slide\lib\async-map.js:47:24)
    313441 verbose stack     at C:\Program Files\nodejs\node_modules\npm\node_modules\read-installed\read-installed.js:263:14
    313441 verbose stack     at cb (C:\Program Files\nodejs\node_modules\npm\node_modules\slide\lib\async-map.js:47:24)
    313441 verbose stack     at C:\Program Files\nodejs\node_modules\npm\node_modules\read-installed\read-installed.js:263:14
    313442 verbose cwd C:\Users\znowden\Desktop\react-slingshot
    313443 error Windows_NT 10.0.10586
    313444 error argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install"
    313445 error node v4.4.3
    313446 error npm  v2.15.1
    313447 error code EPEERINVALID
    313448 error peerinvalid The package react@15.0.1 does not satisfy its siblings' peerDependencies requirements!
    313448 error peerinvalid Peer react-dom@15.0.1 wants react@^15.0.1
[...]

PR #117, opened 2016-04-23
Add babel-register as dev dependency

piotrkaczmarek:
Babel register module was added to [testSetup](https://github.com/coryhouse/react-slingshot/commit/16ec28c9029bf7e2b65b26c22a1c2daadab427a2#diff-6c723dbdc461fd954bf1d26883b049c7R26) but it wasn't added as an explicit dependency. On clean environment `npm test` fails with `Error: Cannot find module 'babel-register'`

The issue was also reported here https://github.com/coryhouse/react-slingshot/issues/116 .


coryhouse:
Thanks!

PR #119, opened 2016-04-26
Rename businessLogic to utils

zsherman:
Maybe it's a personal preference, but I feel like `utils` is a better standard here.


coryhouse:
Hi @zsherman. I chose the name "businessLogic" to convey that it contains logic that isn't tied to React or specific to any front-end library. In my front-end apps, I strive to extract my logic into separate "POJOs" when possible so that it can be easily testing in isolation. I was seeking to convey a "business logic layer" on the client. I'm not sure that "utils" conveys that mission as clearly. 

Given this description, does that change your thoughts on the naming? 

I'm going to let this sit for a bit since I'd love to hear from others on their opinion.

zsherman:
I see what your saying – I don't have a really strong opinion here, the naming just felt a little unconventional compared to "lib" or "utils" or something along those lines.

tsongas:
I agree businessLogic feels icky. Maybe "core" would work?

jondot:
I think 'helpers' is what you guys are looking for.

coryhouse:
Related poll: https://twitter.com/housecor/status/745962536751632385

coryhouse:
@zsherman Thanks for the PR - I just committed this change. Your suggestion of utils ended up being the most popular in the Twitter poll I posted. Thanks everyone for the input!

zsherman:
👍

PR #123, opened 2016-04-28
Indentation cleanup

aminimalanimal:
- applied two-space indentation throughout project
- ensured every file ends with a single new line


coryhouse:
@aminimalanimal Thanks for the PR. I totally agree this needs to be done, but due to the number of impacts and churn that have happened since, I'm handling it manually now.

PR #127, opened 2016-05-02
Remove unnecessary path to node executables

adam-beck:


coryhouse:
I could have sworn there was a good reason I needed to do this, but after testing and reading [Node's docs](https://docs.npmjs.com/cli/run-script) I agree it appears safe to remove.

Thanks for the PR!

coryhouse:
I was able to successfully run a test script without the node_modules path on an old version of Node (0.10.92) using both Mac and Windows so we should be good here.

Issue #128, opened 2016-05-02
How to add favicon to /dist?


jforaker:
Seems trivial but is very difficult.  How to add a `favicon.ico` to the `npm run build` task?

coryhouse:
Great question. Follow the same pattern used by tools/buildHTML. Just copy the file from the root of /src to the root of /dist

jforaker:
Ok so I understand how to create the `<link .. >` html element, but how to "copy" the actual file from /src to /dist?

Here's what I have so far...

```
	const $ = cheerio.load(markup);
	const head = $('head');
	head.children().last()
		.after('<link rel="stylesheet" href="styles.css">')
		.after('<link rel="shortcut icon" href="favicon.ico">');
```

coryhouse:
Ah, multiple ways to do that, but here's an easy way with Node: http://stackoverflow.com/questions/11293857/fastest-way-to-copy-file-in-node-js

jforaker:
Hmm that is one way. Here's how I did it in case anyone lands on this thread:

1. require my favicon in my main App.js entry file 
 `require('../assets/images/favicon.ico')`

2. Added a loader configuration that does not scramble the image file names in the `build` task (per this thread https://github.com/webpack/webpack/issues/1336#issuecomment-171363119)
 ```
		{
			test: /\.jpe?g$|\.ico$|\.gif$|\.png$|\.svg$|\.woff$|\.ttf$|\.wav$|\.mp3$/,
			loader: 'file-loader?name=[name].[ext]'  // <-- retain original file name
		}
```

3. Add the standard `<link rel="shortcut icon" href="favicon.ico">` in src/index.html

Now `npm run build` will output the image/assets names as their original name, hence giving you the ability to  `href="favicon.ico"` right in your index.html

coryhouse:
Nice! That solution is clearly better than my idea. :) I just tried adding this to Slingshot, but I'm getting an error. Would you be willing to submit a pull request? The only change I'd suggest from your snippet is to write a specific test for only the .ico file so that other filetypes can continue to be handled separately.

jforaker:
Sure -->  https://github.com/coryhouse/react-slingshot/pull/129 😄 . Added specific test as you mentioned: https://github.com/coryhouse/react-slingshot/pull/129/files#diff-b83f3305d04bb1c8d674be30b5f9e1efR32  Tested and works well.

chookie:
[...]

PR #129, opened 2016-05-03
Copy favicon.ico asset to the /dist directory during build

jforaker:
Resolves the issue of not being able to "copy" a `favicon.ico` to the `/dist` directory during `npm run build`, see https://github.com/coryhouse/react-slingshot/issues/128.

### DIFF
--- a/webpack.config.prod.js
+++ b/webpack.config.prod.js
@@ -29,6 +29,7 @@ export default {
     loaders: [
       {test: /\.js$/, include: path.join(__dirname, 'src'), loaders: ['babel']},
       {test: /\.(jpe?g|png|gif|svg)$/i, loaders: ['file']},
+      {test: /\.ico$/, loader: 'file-loader?name=[name].[ext]'},
       {
         test: /(\.css|\.scss)$/,
         include: path.join(__dirname, 'src'),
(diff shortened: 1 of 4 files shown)


coryhouse:
Thanks so much! This is slick. Great idea!

webpack.config.prod.js
```diff
@@ -29,6 +29,7 @@ export default {
     loaders: [
       {test: /\.js$/, include: path.join(__dirname, 'src'), loaders: ['babel']},
       {test: /\.(jpe?g|png|gif|svg)$/i, loaders: ['file']},
+      {test: /\.ico$/, loader: 'file-loader?name=[name].[ext]'},
```
andela-kadeniyi: @jforaker I will suggest that you `{test: /\.ico$/, loader: 'file?name=[name].[ext]'}` instead since the latest webpack complains about adding the loader extension to loaders.

kwelch: I would also recommend this from a consistency stance.

andela-kadeniyi: @kwelch Do you mind if I create a PR for this?

kwelch: yeah, that works.

jforaker: please do @andela-kadeniyi

andela-kadeniyi: @jforaker @kwelch Looks like this is no longer needed since the master branch has been updated to webpack 2.

kwelch: Agreed, master looks good.

PR #132, opened 2016-05-09
Don't use no-unused-vars:0 and jsx-uses-vars:1 together

roburidge:
We don't need to enable "jsx-uses-vars" when "no-unused-vars" is disabled. Linting for unused vars is useful so I've left no-unused-vars enabled from "eslint:recommended", however disabling "no-unused-vars" and removing "jsx-uses-vars" would be equally valid.

"If you are not using JSX or if you do not use the no-unused-vars rule then you can disable this rule." ref: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-uses-vars.md#when-not-to-use-it

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -30,7 +30,6 @@
     "semi": [1, "always"],
     "no-trailing-spaces": 0,
     "eol-last": 0,
-    "no-unused-vars": 0,
     "no-underscore-dangle": 0,
     "no-alert": 0,
     "no-lone-blocks": 0,


coryhouse:
I'm confused. You removed no-unused-vars, but left jsx-uses-vars enabled. Given the link you provided, it appears we should either have both enabled or both disabled. Am I missing something?

roburidge:
I'm sorry, I should have mentioned that the "extends": "eslint:recommended" property enables "no-unused-vars": http://eslint.org/docs/rules/#variables

coryhouse:
Ah, great point. Thanks for the PR! 👍

PR #137, opened 2016-05-12
Escape dot in webpack file extension regex match

rndstr:


coryhouse:
Good catch. Thanks for the PR! 👍

PR #140, opened 2016-05-13
Added .scss as excluded feature for tests

rndstr:
While `.scss` is currently nowhere used in files that are imported from tests, it is possible that someone might `import` sass files directly from components which then will fail for tests.

I thought about adding `.ico` but since this will only ever be used in `index.js` I don't see why anyone would have tests there. I'm happy to add it if anyone thinks it's necessary, though.


coryhouse:
Great idea! I like your refactor as well. Nice work. 👍

PR #143, opened 2016-05-17
Absolute styles.css path for multi-level paths

rndstr:
Same as `/bundle.js` to support entry points for clients at example.com/en/something. Otherwise you get 404s on production.

Also see #79


coryhouse:
Good call. Thanks for the PR!

PR #144, opened 2016-05-17
Load styles last for same beahviour as in dev mode

rndstr:
In dev mode, local styles are loaded *after* other CSS included in `<head>`. For prod, the `prepend()` makes `styles.css` be parsed *before* the other CSS in`<head>`.

One fix is to use `append()` instead to make sure local styles are loaded last and in the same order as in development.


coryhouse:
Great point. 👍  This makes sense to me, but can you please do me a favor and get latest to resolve a merge conflict and resubmit? Unfortunately someone else's separate PR changed the same line for a diff reason yesterday. :)

rndstr:
@coryhouse that was me :) rebased, thank you!

coryhouse:
Thanks @rndstr! :)

PR #154, opened 2016-06-04
Add babel's react-html-attrs plugin

sanchan:
This way instead of doing this:

`<div className="my-class"></div>`

we can do:

`<div class="my-class"></div>`

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,6 +3,9 @@
     "es2015",
     "react"
   ],
+  "plugins": [
+    "react-html-attrs"
+  ],
   "env": {
     "development": {
       "presets": [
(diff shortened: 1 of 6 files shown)


coryhouse:
@sanchan Thanks for the PR! I agree this is handy for saving a few keystrokes, however, I'm going to pass on this for two reasons:

1. I want to avoid creating confusion. Those unaware of this plugin will look at the example JSX and think it has syntax errors.
2. I want to keep this kit as simple as I can. 

That said, thanks again for the PR - I wasn't even aware this existed, and may try it on my next project. :)

sanchan:
@coryhouse ok, I understand :)

Thanks for consider my PR and I'm glad you know about it, I recommend it for your next project for sure :D

Have a good week sir!

PR #156, opened 2016-06-05
Fix travis sending coverage to coveralls

Oliboy50:
## Why
As you can see here : https://travis-ci.org/coryhouse/react-slingshot/jobs/135371065#L1104
Travis currently fails to send build info to coveralls.
So this PR fixes this (tried on a project "nodejs 6" right now)


coryhouse:
Great catch. Thanks so much!

PR #160, opened 2016-06-07
Setup script

gargrave:
Setup script to perform the following actions:

- Install npm packages
- Remove the original Git repo
- Prompt the user to update `package.json` where appriopriate
- Remove the `tools/setup` folder containing the script when finished

I did add two devDependencies in `package.json`: `prompt` and `replace`, for handling the `package.json` updates. If you would prefer that I not save them as dependencies, I could look at having the script just manually install then uninstall them.

Let me know if you have any questions or issues.

Thanks,
Gabe

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ Why Slingshot?
 
 ## Get Started
 1. **Initial Machine Setup**. First time running the starter kit? Then complete the [Initial Machine Setup](https://github.com/coryhouse/react-slingshot#initial-machine-setup).
-2. **Clone the project**. `git clone https://github.com/coryhouse/react-slingshot.git`.  
-3. **Install Node packages**. `npm install`
+2. **Clone the project**. `git clone https://github.com/coryhouse/react-slingshot.git`.
+3. **Run the setup script**. `npm run setup`
 4. **Run the example app**. `npm start -s`
 This will run the automated build process, start up a webserver, and open the application in your default browser. When doing development with this kit, this command will continue watching all your files. Every time you hit save the code is rebuilt, linting runs, and tests run automatically. Note: The -s flag is optional. It enables silent mode which suppresses unnecessary messages during the build.
 5. **Review the example app.** This starter kit includes a working example app that calculates fuel savings. Note how all source code is placed under /src. Tests are placed alongside the file under test. The final built app is placed under /dist. These are the files you run in production.
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "4.0.0",
   "description": "Starter kit for creating apps with React and Redux",
   "scripts": {
+    "setup": "node tools/setup/setupMessage.js && npm install && node tools/setup/setup.js",
     "remove-demo": "babel-node tools/removeDemo.js",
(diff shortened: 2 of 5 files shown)


coryhouse:
This is awesome! Brilliant work! 💯 

I'm making a few tweaks and will get this merged today. Thanks so much! You nailed it! 👍

coryhouse:
I made some tweaks:

1. Converted to ES6
2. Fixed linting issues
3. Tweaked messaging to use chalkConfig

I'm torn on whether to remove the git repo. It actually bit me as I was trying your PR. I had made some tweaks and then wiping out the repo removed my changes. 

I'm also trying to remove the keywords from package.json, but am struggling with the proper regex (not my strongsuit). Certainly open to a suggestion there! 

Thanks again!

gargrave:
Yeah, I feel you on the Git issue--I can safely say I have never had to try to deal with writing a script that deletes Git _and itself_--it certainly made for an interesting time trying to keep it tracked. It seems like the only time it would really be an issue is for someone who is planning on contributing, and thus needs to keep the original repo--otherwise, I think it's a really handy feature to have in there. (Maybe just adding a cautionary note to the relevant line in the README file would be enough?)

Keywords: I started looking at that as well, and felt like it was probably more effort than it was worth in that case--mostly, like you said, because it doesn't use the same simple regex that the other prompts do. Are you thinking you would like to prompt the user for new keywords, or just clean it out and leave an empty array for the user to populate manually?

coryhouse:
I was just going to clean out the existing keywords, or, if easier, remove the keywords section altogether.

gargrave:
Cool, that should be easy enough. I think I have it already, but I just need to test a full run of it when I have a minute. I'll send another PR later today.

gargrave:
Hey Cory,

One quick thing I noticed in trying to re-clone from scratch and run this. Maybe it's something wrong on my end, but when I am starting from a clean clone of the repo, the script won't run using `babel-node`. I ran into this before, and it's why I was originally sticking with mostly ES5 syntax that "vanilla Node" will run, since the npm script itself is actually launched before the babel tools are installed locally.

Am I maybe missing a global install that is causing this, or do we need to switch it back to the original ES5-ish syntax, since the assumption will be that anyone running this is starting from a clean clone with no packages installed?

I do fully admit that this could be user error on my part, but I just wanted to check before proceeding with anything.

Thanks again!

coryhouse:
Ah! Great point. I didn't consider that. Since babel-node isn't installed until npm install is run, it indeed needs to be in ES5. If your next PR has it back in ES5, I'll just merge it in to complete the rollback of my mistake. 👍

gargrave:
Ok, cool. I will send a PR later on today. (Incidentally, this is the same reason I didn't use `chalk` in the initial setupMessage.js file.)

coryhouse:
Yep, that makes sense. 

I also realized we need a bit of validation on the project name. No spaces are allowed.

gargrave:
Yeah, I had considered that possibility as well. It's actually pretty trivial to add regex checking to user input with `prompt`, so it shouldn't be too big of a deal to enforce npm's naming conventions. Pretty confident I can work it in!

PR #161, opened 2016-06-08
Updated setup script

gargrave:
Updated setup tools with the following changes:

- Reverted back to the ES5-ish syntax, since we don't have the babel packages installed yet when the script is launched (I stayed as faithful as possible to the updates you made to my original version)
- Replace the existing 'keywords' array with an empty one
- Validate the name of the package according to npm requirements.

Regarding the validation, I referenced the two links below to try to get the right regex--everything seems solid in my testing, but I am happy to re-visit if you spot any holes. (Basically: URL-friendly, and cannot start with a period or underscore.)

https://github.com/npm/npm/blob/master/doc/files/package.json.md#name
https://stackoverflow.com/questions/695438/safe-characters-for-friendly-url

Let me know if you have any questions or issues.

Thanks,
Gabe

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "webpack-dev-middleware": "1.6.1",
     "webpack-hot-middleware": "2.10.0"
   },
-  "keywords:": [
+  "keywords": [
     "react",
     "reactjs",
     "react-router",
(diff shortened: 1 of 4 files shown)


coryhouse:
Looks great! Thanks so much for the contribution and quick turnaround on the tweaks! 👍

PR #166, opened 2016-06-15
Remove typo from comment

fellipeale:
Just removed the 'new webpack.HotModuleReplacementPlugin());' from the comment block


coryhouse:
Good catch. Thanks! :+1:

PR #172, opened 2016-06-20
Add [hash] to the names of css and js bundles.

vangelov:


coryhouse:
Thanks @vangelov! However, after digging into this more, I think there's a piece missing.

Here's the issue:
We should only change the bundle's filename when code changes. This way users only download new JS / CSS when something changes. This is possible, but requires [some extra work](https://github.com/okonet/webpack-long-term-cache-demo/blob/master/webpack.config.js).

[Here's the related blog post](https://medium.com/@okonetchnikov/long-term-caching-of-static-assets-with-webpack-1ecb139adb95#.6r8rkbpkh).

And here's where I'm torn: Is the complexity worth it? This extra complexity only benefits people who:
1: Want to set their assets to never expire for the ultimate in performance
2: Setup the appropriate HTTP headers on their webserver

The current static filenames work just fine. Assuming the webserver is configured to properly utilize etags, utilizes very little extra bandwidth anyway. 

Bottom-line, currently torn on whether the benefit justifies the cost in this case.

What do you think?

vangelov:
I think you're right. It does complicate things. 

My main concern was to be able to invalidate the browser cache in case I need to deploy an emergency bug fix, for example, but you can always change the name of the files if it comes to that.

coryhouse:
@vangelov I just committed a change that handles this. After doing some digging, I found html-webpack-plugin is a slick way to get this done. Now, files will receive a new hash as they change and index.html is generated dynamically with the appropriate references. I'd love to hear what you think. Thanks for the inspiration!

PR #173, opened 2016-06-20
update: babel-node scripts to execute from the .bin executables

FarhadG:
Best practices for referencing the executables from the `.bin` directory.


coryhouse:
@FarhadG - I see you opened this then closed it. I had a question about why referencing .bin explicitly was necessary (to my knowledge, it shouldn't be...)

FarhadG:
I closed it and re-opened it so that I could match the repos `node` environment. Since `babel-node` may or may not be installed globally (e.g. developers don't want these global commands), it's better to include them as `dependencies` (as they are) and execute the commands from the `.bin` directory.

New PR: https://github.com/coryhouse/react-slingshot/pull/174

PR #174, opened 2016-06-20
update: babel-node scripts to execute from the .bin executables

FarhadG:
Best practices of using executables from the `.bin` directory.


coryhouse:
Do you have a link that explains why this is a best practice? I've found it unnecessary.

FarhadG:
@coryhouse Since `babel-node` may or may not be installed globally (e.g. developers don't want these global commands), it's better to include them as `dependencies` (as they are) and execute the commands from the `.bin` directory.

coryhouse:
I'm confused - You don't have to install the dependencies globally for the current setup to work. 

When evaluating scripts, npm adds node_modules/.bin/ to the PATH environment variable. That means you can forego the explicit .bin folder path is unnecessary. [More info](https://firstdoit.com/no-need-for-globals-using-npm-dependencies-in-npm-scripts-3dfb478908#.14c84hql5)

FarhadG:
That's awesome! I guess I'm still old-school with the older versions of `npm`

PR #179, opened 2016-06-29
Re-Add replace dependency

w3cj:
tools/setup/setup.js uses this and is currently breaking because of the missing dependency
https://github.com/coryhouse/react-slingshot/blob/master/tools/setup/setup.js#L4


coryhouse:
Good catch. Thanks!

PR #182, opened 2016-07-08
Merge dev and prod to index in store

ngtan:
Something I see an error "No default export found in module import/default" when I import configureStore into app/index.js
but the application is still running.  So, I think we should merge config dev and config store code into index file.
And another problem, if have dev and prod file, we must code in 2 file. I think I spent a lot of time for the problem.


coryhouse:
@ngtan Good points. Thanks for the pull request!

coryhouse:
Whoops, correction. I just rolled this back. The reason these are separate files is to assure that no dev specific code is included in the production bundle. With this pull request, development specific code like thunk/saga/reduxImmutableStateInvariant, etc would be included in the production bundle. Why? Because ES6 imports are statically analyzable. So the import at the top of a single configureStore.js file would have to include development-specific resources that we don't want in production.

The real solution to your problem: Make sure index.js has this on line 1 (as reflected in the repo):
/* eslint-disable import/default */

ngtan:
I got it. Thanks Cory.

PR #184, opened 2016-07-08
Clarifying name and use of React env config.

HoraceShmorace:
1. "GLOBALS" is ambiguous, particularly since its only use is to configure the React environment. I don't really see a reason to abstract this at all, as opposed to just inlining the object in the call to `webpack.DefinePlugin`.

2. Clarified the comment on line 25 to indicate that the statement before it sets the build to either mode, not just prod.

Note: There should probably be a comment on line 6 to explain the need to `JSON.stringify` a string. I'm not sure of the reason, so I couldn't add the comment myself.

### DIFF
--- a/webpack.config.dev.js
+++ b/webpack.config.dev.js
@@ -2,7 +2,7 @@ import webpack from 'webpack';
 import path from 'path';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 
-const GLOBALS = {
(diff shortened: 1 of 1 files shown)


coryhouse:
Thanks for the PR!

coryhouse:
And I just inlined the code as you suggested. Thanks!

PR #202, opened 2016-07-15
added stage1 babelrc settings

ArcanisCz:
According to https://babeljs.io/docs/plugins/preset-stage-1/, which is included in dev-dependencies, you need to specify stage1 in .babelrc to work. I assume, you only forgot this, since it is already included.

After adding, features like {var1, var2, ...other} started to work.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,8 @@
 {
   "presets": [
     "es2015",
-    "react"
+    "react",
+    "stage-1"
   ],
   "env": {
     "development": {


coryhouse:
Interesting - I had intended to remove stage 1 support because I didn't think ESLint could handle stage 1-3 features, but object spread is the most relevant feature and is already enabled in .eslintrc, so I went ahead and accepted. The one caveat with this change is some other stage 1-3 features may cause ESLint warnings. But I figure those who experience that can either disable eslint on that line, or pull in babel-eslint to have full stage 1-3 support.

Thanks for the PR!

ArcanisCz:
If you want only object spread without other stage 1-3 features, why not include only object spread instead of stage 1-3 bundle? (http://babeljs.io/docs/plugins/transform-object-rest-spread/ - this one is responsible for it)

then, you will add only `{
  "plugins": ["transform-object-rest-spread"]
}`

coryhouse:
Yeah, I considered it, but figure supporting all of stage-1 is more inclusive and only downside is potential linting issues which people can easily address.

Issue #205, opened 2016-07-18
Browser Sync won't load bundle.js on external devices for dev build


SunnyJohal:
Node version: 6.2.2

npm version: 3.9.5

Operating system: MAC OSX El Capitan (v 10.11.15)

Command line used: npm start -s

**Steps to reproduce:** 

- Npm install the starter kit
- Run the command in dev mode: npm start -s

- Once the app is loaded locally: On an external device enter the External access URL in the browser.

- Open the browser console, you will see the error message: http://localhost:3000/bundle.js net::ERR_CONNECTION_REFUSED

**What I've tried:**

- Using a different port for the app (no success).

- Trying a different external device (another laptop and phone, no luck).

- Using a different machine to run the app on (another mac, no luck).

**Note:** 

Browser sync hot reloading works fine with the production build. It's the development environment that I appear to be having issues with .

Really love this starter kit and want to use it so hope it's an issue that I can get resolved. Cheers

Sunny

kwelch:
Does the external address work at all? If you use it on the local device with the external address does it load?

SunnyJohal:
Yes the external address works on my local device. Thanks for your response.

coryhouse:
This is interesting timing because I was considering removing browsersync since it complicates configuration and I suspect few are using its sync features.

That said, great catch. I've reproduced the issue. I don't have a fix idea at this point, but thanks for the heads up.

SunnyJohal:
I have a feeling that it is something to do with the webpack-dev-middleware package. I'll let you know if I find a fix for this issue. Cheers

SunnyJohal:
Ok when I edit the `publicPath` property in `webpack.config.dev.js` to use the ip address on my local machine (in place of localhost) then hot reloading works. Not ideal but a temporary fix for anyone reading this issue.

coryhouse:
Yep, you're on the right track. That's where I was focused as well. I hoped removing the publicPath altogether would resolve it, but that didn't fix it for me. 

It does appear we don't need publicPath anymore. I could have sworn it was necessary before.

SunnyJohal:
An even better fix is to simply comment out the `publicPath` property in `webpack.config.js`

coryhouse:
Ah, indeed! Problem is, the public path was added to resolve this issue in dev: https://github.com/coryhouse/react-slingshot/pull/125

Any ideas on how to support both use cases?

SunnyJohal:
[...]

Issue #181, opened 2016-07-02
npm start wants bundle.js on port 3000, while website started on first free port


svdoever:
Node version: 6.2.1

npm version: 3.9.3

Operating system: OSX

Command line used: npm start

Steps to reproduce: **npm start** seems to do everything, but when site starts up gives error in browser console window **Failed to load resource: the server responded with a status of 404 (Not Found) **http://localhost:3000/bundle.js**, but the site started at port 3004! Seems like link to **bundle.js** is hard-coded on port **3000**, while site starts up on next free port. I had another programming running that took up port 3000...

coryhouse:
You can change the port by updating srcServer.js, distServer.js, and webpack.config.dev.js.

I haven't found a way to run at a dynamic port, but would love to see a pull request that handles it.

svdoever:
Hi Cory, no problem... now I know I have to kill other apps taking port 3000. Might be good to take a completely different port i.e. 3918 or something like that, because most apps take port 3000:-) 

By the way: my compliments on your course on PluralSight!

coryhouse:
Resolved with https://github.com/coryhouse/react-slingshot/commit/092151e6489b1d2c03bf0420fb3c84c5de50dc41

PR #207, opened 2016-07-21
Support bootstrap glyiphicons without hardcoding public path

kwelch:
So this removes the public path explicitly set causing issues (#181 & #205)

This is a very hacky solution in my mind. 

_Special Notes_
- Adding bootstrap to dependancies => this was intention since I didn't expect the bootstrap part to "live"
- ~~I was able to run on external device, but I was not seeing bootstrap icons~~ re-tested and this works

### DIFF
--- a/src/bootstrap-public-path.js
+++ b/src/bootstrap-public-path.js
@@ -1,0 +1,4 @@
+/**
+ * Created by welchk on 7/20/16.
+ */
+__webpack_public_path__ = window.location.protocol + "//" + window.location.host + "/";
--- a/src/bootstrap-public-path.js
+++ b/src/bootstrap-public-path.js
@@ -1,4 +1,0 @@
-/**
- * Created by welchk on 7/20/16.
- */
-__webpack_public_path__ = window.location.protocol + "//" + window.location.host + "/";
--- a/src/components/NotFoundPage.js
+++ b/src/components/NotFoundPage.js
@@ -5,7 +5,7 @@ const NotFoundPage = () => {
   return (
     <div>
       <h4>
-        404 Page Not Found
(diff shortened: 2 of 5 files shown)


src/bootstrap-public-path.js
```diff
@@ -0,0 +1,4 @@
+/**
+ * Created by welchk on 7/20/16.
+ */
+__webpack_public_path__ = window.location.protocol + "//" + window.location.host + "/";
```
coryhouse: Where is this variable used?

kwelch: This is dynamically sets the public path at runtime allowing proper resolving for loaders. 

Found it in the issue for relative paths in `style-loader`. 
Further explaination: https://github.com/webpack/style-loader/pull/96#issuecomment-228370602

Honestly, I am not a fan of this outside of the fact that it does work. This will allow proper support icon support to fix #125, but it also does so in a way that does not cause other breaking changes as seen in #181 & #205.

kwelch:
I can revert the bootstrap specific changes if you want to move forward with this.

src/bootstrap-public-path.js
coryhouse: Ah, I see that [__webpack_public_path__ is a magic global](https://github.com/webpack/webpack/issues/443).

I'd suggest renaming bootstrap-public-path to webpack-public-path and adding a comment inside that explains why its being set dynamically and includes a link to relevant issues and the docs that explain what  __webpack_public_path__ is.

Thanks so much for all your hard work on this Kyle! 💯

coryhouse:
Great work Kyle! Thanks for all your help on this! 👍

Issue #162, opened 2016-06-13
Test coverage includes `spec` files in its report


FarhadG:
I've noticed that the test coverage includes the `spec.js` files in its report, which, I assume, is not intended behavior as it conflates the overall report.

Node version: 5^
npm version: 3^
Operating system: OSX
Command line used: npm run test:cover

coryhouse:
@FarhadG - Great point. Do you happen to know an easy way to exclude them?

FarhadG:
Not at the moment, @coryhouse. I'll take a look, though.

coryhouse:
I haven't figured out a way to exclude them from the report yet, but would welcome insight from anyone willing.

peel3r:
how about moving all tests to "test" directory?

peel3r:
To fix this  create a .istanbul.yml file in the root of react-slingshot with:
`instrumentation:
  excludes: ['*.spec.js']
  extensions: ['.js']`

coryhouse:
Awesome! That was easy. Thanks for the tip! 👍 

Commit: https://github.com/coryhouse/react-slingshot/commit/492e5b7e92740b1e80dd07ed65a18d7d07ab0fd9

PR #213, opened 2016-07-26
Exclude test files in Coverage

kwelch:
Implemented @peel3r 's recommendation to exclude the spec files.  https://github.com/coryhouse/react-slingshot/issues/162#issuecomment-235104596


coryhouse:
Ha! Literally at nearly the same moment. :) I think I beat you by 10 seconds?!

Sorry about that, but thanks Kyle!

PR #221, opened 2016-08-04
Move chunk injection into template

kwelch:
I came across an issue in a different project that uses `html-webpack-plugin` that with the `inject` flag set to `true` the path to bundle is relative and causes issues with nested paths. 

I tested this in slingshot by navigating to http://localhost:3000/demo/test, I expected to see the 404 page, but with the relative path to `bundle.js` it returned a blank page. 

I used [this example template](https://github.com/jaketrent/html-webpack-template/blob/86f285d5c790a6c15263f5cc50fd666d51f974fd/index.html) to inject the chunks and properly made them absolute.

### DIFF
--- a/webpack.config.dev.js
+++ b/webpack.config.dev.js
@@ -31,7 +31,8 @@ export default {
         removeComments: true,
         collapseWhitespace: true
       },
-      inject: true
+      // injection is now handled in the template to allow for absolute paths, fixes issue with nested routes
+      inject: false
     })
   ],
   module: {
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -18,8 +18,15 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta charset="utf-8"/>
   <title>Fusion Starter</title>
+  <% for (var css in htmlWebpackPlugin.files.css) { %>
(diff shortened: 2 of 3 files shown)


kwelch:
I found an issue with this running a build. The chucks have the webpack config publicPath prepended and when running from src absolute paths are fine, but the resulting files in the dist directory have the `/` twice which causes it to fail to load when served. 

I will update PR.

webpack.config.dev.js
```diff
@@ -31,7 +31,8 @@ export default {
         removeComments: true,
         collapseWhitespace: true
       },
-      inject: true
+      // injection is now handled in the template to allow for absolute paths, fixes issue with nested routes
```
coryhouse: I'm concerned that setting inject to false disables **all** injection, which means we have to provide manual injections for all potential file types. For example, would Bootstrap related filetypes need to be explicitly injected in the .ejs file with this setup?

kwelch: I completely understand. I am running this setup with bootstrap but I am using the bootstrap-loader. 

I want to try one another approach on this after sleeping on these changes and seeing a bug in the production build. I will try and find an approach that does not mean taking on injection.

kwelch:
Closing #227 properly addressed this issue without having to manually handle injection.

PR #226, opened 2016-08-05
Fix needed path dependency.

iLeonelPerea:
@coryhouse 

We need path dependency in webpack...js.

Tnks


coryhouse:
I refactored out the need for it in the previous commit. Nothing is calling path anymore and it runs fine. Am I missing something?

PR #227, opened 2016-08-07
Fixes bundle.js not loading on nested page refresh

endlesscombo:
1.  Given a nested route: `<Route path="course/:id" component={ManageCoursePage} />`
2.  If the user tries to go directly to `http://localhost:3000/course/4` in their browser.
3.  bundle.js will attempt to load from `http://localhost:3000/course/bundle.js` and get a 404 error.

Found this fix here:  https://github.com/ampedandwired/html-webpack-plugin/issues/98

If you would like an example project with this behavior, please let me know and I will post one.

### DIFF
--- a/webpack.config.dev.js
+++ b/webpack.config.dev.js
@@ -15,6 +15,7 @@ export default {
   target: 'web', // necessary per https://webpack.github.io/docs/testing.html#compile-and-test
   output: {
     path: `${__dirname}/src`, // Note: Physical files are only output by the production build task `npm run build`.
+    publicPath: '/',
     filename: 'bundle.js'
   },
   plugins: [


coryhouse:
@kwelch recently committed [this change](https://github.com/endlesscombo/react-slingshot/commit/2f6f679925bfc8b88307bc5dfe7ad9206acbfee5) that dynamically sets the publicPath. The suggested change in this PR would override his work and cause https://github.com/coryhouse/react-slingshot/pull/125 and https://github.com/coryhouse/react-slingshot/issues/205 to recur again.

@kwelch - Is there a tweak we could make to your work to support this?

kwelch:
Yes, #221 addresses this issue, but had a bug for production. I am pretty sure this works though, I had been playing with a solution like this on Friday that used public path and set it dynamically, and it fixed both background images and nested routes. 

I can look further at this specific PR tomorrow, but I am quite sure it will work.

kwelch:
I check ran some tests on this checking the proper injection and regression points with dynamically setting public path and all looked good. 

Thanks for the PR!

endlesscombo:
Thanks to both of you!  This project and the video course are really helping me make sense of the react ecosystem.

Issue #121, opened 2016-04-26
Include redux-thunk by default


zsherman:
Would be nice to have an example of this and the middleware all wired up.

coryhouse:
Considering it, but Redux thunk is easy to wire up and one of 3 popular options. Not sure having an opinion on this is worth it.

offsky:
Im curious. Can you please let us know the other 2 popular options?  Ive been happy with thunk, but Id like to know if there is something better.

Here is how to add redux-thunk. If you don't want to include it by default in this project, maybe this setup could be added to the FAQ or as a separate option for those people who want it.  Or maybe this issue left open is enough for people to find.  

Thanks for making react-slingshot! It saved me a bunch of time bootstrapping my new project.

configureStore.prod.js
```
import {createStore} from 'redux';
import thunk from 'redux-thunk';
import rootReducer from '../reducers';

export default function configureStore(initialState) {
  return createStore(rootReducer, initialState, applyMiddleware(thunk));
}
```

configureStore.dev.js
```
import {createStore, compose, applyMiddleware} from 'redux';
import thunk from 'redux-thunk';
import rootReducer from '../reducers';

export default function configureStore(initialState) {
  const store = createStore(rootReducer, initialState, compose(
        applyMiddleware(thunk),
        window.devToolsExtension ? window.devToolsExtension() : f => f
    )
  );
...
```

Then in an action that wants to use thunk
```
export function myAsyncAction(foo) {
	return function (dispatch, getState) {
		setTimeout(function() {
			dispatch({type: types.MY_ACTION, foo});
		},5000);
	};
}


```

coryhouse:
Given all the options, I'm not sure having an opinion makes sense, but I'm open to accepting a pull request using Redux thunk.

From the [Redux docs](http://redux.js.org/docs/advanced/AsyncActions.html):

Thunk middleware isn’t the only way to orchestrate asynchronous actions in Redux:

You can use redux-promise or redux-promise-middleware to dispatch Promises instead of functions.
You can use redux-observable to dispatch Observables.
You can use the redux-saga middleware to build more complex asynchronous actions.
You can even write a custom middleware to describe calls to your API, like the real world example does.
It is up to you to try a few options, choose a convention you like, and follow it, whether with, or without the middleware.

kwelch:
[...]

PR #232, opened 2016-08-12
Include Redux Thunk Middleware

kwelch:
Based on the conversation in #121, I added the thunk middleware with proper documentation and updated the save action to be a thunk as a demo of both what it looks like and how to test. 

Let me know if the documentation is sufficient and clear.


kwelch:
Should be able to remove the settings argument also from the actions since it can be accessed from the thunk, that being said I don't see it being used prior to this change. I am not sure its use so I didn't want to touch it.

coryhouse:
Thanks for the PR @kwelch! 👍

PR #233, opened 2016-08-12
Proposal: Include Selectors

kwelch:
I think selectors are really awesome for separating state from components and preform calculations. 

I started on this effort since I saw both a need from the app, and I think it is a good practice to not store calculated values in the store. 

That being said, some of the logic in the app counter acts these changes, so I paused the effort to make sure this goes in the right direction. Additionally, these changes may benefit from including `redux-thunk` (#232).

I wanted it to be an issue, but as a PR it is simpler to compare.

### DIFF
--- a/src/containers/FuelSavingsPage.js
+++ b/src/containers/FuelSavingsPage.js
@@ -1,8 +1,9 @@
-import React, {PropTypes} from 'react';
-import {connect} from 'react-redux';
-import {bindActionCreators} from 'redux';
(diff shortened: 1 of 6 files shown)


coryhouse:
While I agree that selectors are a useful pattern, I don't believe they're universally necessary, so I'm not sold that they make sense in the kit. It's yet another thing to understand and learn when trying this out, so I lean toward omitting.

kwelch:
I understand that and agree. Closing.

PR #236, opened 2016-08-16
[Proposal] Implement Css Modules

kwelch:
I have added Css Modules, with a working example. I have seen a few request for it. I also verified that with commit it will work with react-toolbox no changes necessary. 

@coryhouse, I am on the fence on actually including merging this. I would be fine leaving it as a how-to implement.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -66,6 +66,12 @@
     "react/sort-comp": 1,
     "react/wrap-multilines": 1
   },
+  "settings": {
+    "import/ignore": [
+      "node_modules",
+      "\\.(coffee|scss|css|less|hbs|svg|json)$",
+    ]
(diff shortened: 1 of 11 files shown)


coryhouse:
Since this is a **React** starter, I'd vote to leave this out. I added Sass-loader initially because it's very popular but in retrospect I arguably shouldn't have taken an opinion on styling at all. There are literally over a dozen ways to handle CSS and I don't see a clear winner at this point.

kwelch:
I agree. I am a fan of it because it put styling at the component level, but I agree I would not call it the clear winner yet. I am interested in seeing how libraries like aphrodite land, because it so easily translate to what you expect on the React-Native side. 

I will close.

PR #238, opened 2016-08-16
Add support for .jsx extension.

jboxman:
Simple change to allow .jsx extension to be used. Using it right now in both development and production modes.

Thanks!


kwelch:
I am not against using support for `jsx`, but I do agree with this [comment](https://github.com/airbnb/javascript/pull/985#issuecomment-239145468)

coryhouse:
Thanks @jboxman! Looks great!

coryhouse:
@kwelch I agree and always use .js, but understand some people's preference. In this case, a small price to pay to be more inclusive. And hey, got @jboxman his first PR! 👍

Issue #240, opened 2016-08-18
Reducers should be pure functions


samit4me:
In the example reducer *fuelSavingsReducer.js*, it is updating the property **dateModified** with a `new Date()`. 

This does not prevent it from working, but it will give some strange results when using the time travelling feature of Redux (e.g dateModified will be different).

> Things you should never do inside a reducer:
> * Mutate its arguments;
> * Perform side effects like API calls and routing transitions;
> * Call non-pure functions, e.g. Date.now() or Math.random().

http://redux.js.org/docs/basics/Reducers.html#handling-actions

A possible solution would be to pass the **dateModified** as an argument from the action creator `saveFuelSavings()` something like this:

```Javascript
dispatch({
  type: types.SAVE_FUEL_SAVINGS,
  dateModified: dateHelper.getFormattedDateTime(new Date()),
  settings,
})
```

coryhouse:
Excellent point! Would you be willing to submit a pull request?

samit4me:
Sure no problems

coryhouse:
Nothing formal required - branch and then submit the PR. Thanks in advance!

PR #241, opened 2016-08-19
#240 - Make reducers pure by passing date as an argument.

samit4me:
Reducer is now pure because the Date is passed as an argument.

This change makes testing the ActionCreators more difficult as the date needs to be mocked, so the date is being mocked using [mockdate](https://github.com/boblauer/MockDate). I've tried to mock using [babel-plugin-rewire](https://github.com/speedskater/babel-plugin-rewire) but ran into issues regarding named exports which seem to be related to [this issue](https://github.com/speedskater/babel-plugin-rewire/issues/109).

### DIFF
--- a/src/reducers/fuelSavingsReducer.spec.js
+++ b/src/reducers/fuelSavingsReducer.spec.js
@@ -41,6 +41,7 @@ describe('Reducers::FuelSavings', () => {
       }
     };
   };
+  const dateModified = dateHelper.getFormattedDateTime(new Date());
 
   it('should set initial state by default', () => {
     const action = { type: 'unknown' };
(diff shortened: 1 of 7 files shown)


src/reducers/fuelSavingsReducer.spec.js
```diff
@@ -41,6 +41,7 @@ describe('Reducers::FuelSavings', () => {
       }
     };
   };
+  const dateModified = dateHelper.getFormattedDateTime(new Date());
```
coryhouse: Passing new Date() is unnecessary here since it mirrors the default param

coryhouse:
Thanks for the PR @samit4me ! Great work! 👍

samit4me:
Wow that was quick, thanks.

kwelch:
@samit4me Congrats on the PR! 👍

PR #243, opened 2016-08-22
activate redux-immutable-state-invariant

sheldonwjones:
Utilize redux-immutable-state-invariant to help with development.

https://github.com/coryhouse/react-slingshot/blob/master/package.json#L79

### DIFF
--- a/src/store/configureStore.dev.js
+++ b/src/store/configureStore.dev.js
@@ -3,12 +3,16 @@
 // With Redux, the actual stores are in /reducers.
 
 import {createStore, compose, applyMiddleware} from 'redux';
+import reduxImmutableStaeInvariant from 'redux-immutable-state-invariant'
 import thunkMiddleware from 'redux-thunk';
 import rootReducer from '../reducers';
 
 export default function configureStore(initialState) {
   const middewares = [
     // Add other middleware on this line...
+
+    // Redux middleware that spits an error on you when you try to mutate your state either inside a dispatch or between dispatches.
+    reduxImmutableStaeInvariant(),
 
     // thunk middleware can also accept an extra argument to be passed to each thunk action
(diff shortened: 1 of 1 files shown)


src/store/configureStore.dev.js
```diff
@@ -3,13 +3,17 @@
 // With Redux, the actual stores are in /reducers.
 
 import {createStore, compose, applyMiddleware} from 'redux';
+import reduxImmutableStaeInvariant from 'redux-immutable-state-invariant'
```
coryhouse: There's a typo in the variable name. :)

sheldonwjones:
Sorry about that. I copied that line from a project that has src/.eslintrc set to.
`
{
  "rules": {
    "semi": [2, "never"]
  }
}
`

src/store/configureStore.dev.js
```diff
@@ -3,13 +3,17 @@
 // With Redux, the actual stores are in /reducers.
 
 import {createStore, compose, applyMiddleware} from 'redux';
+import reduxImmutableStaeInvariant from 'redux-immutable-state-invariant';
```
coryhouse: reduxImmutableStaeInvariant is missing a "t". Should be reduxImmutableStateInvariant.

:)

coryhouse:
@kwelch Thoughts on this? Worth accepting?

kwelch:
I had never seen this before. I am for it, very useful for a new developer, especially if they are not use to immutability & pure functions. 

It even caught some errors that happen to me when pulling data from state.  💯

sheldonwjones:
A better name might be immutableStateInvariantMiddleware or even reduxImmutableStateInvariantMiddleware. I'm not sure the best practice here.

Thoughts?

coryhouse:
Thanks @sheldonwjones ! 👍

PR #258, opened 2016-09-11
Fix node version check

alucic:
This PR

* [x] removes unused `stderr` param
* [x] removes unreachable `process.exit(1)`
* [x] throw error on unmet node version requirement 

>A floating point number parsed from the given string. If the first character cannot be converted to a number, NaN is returned.

`node -v` returns version number prefixed with `v`, which makes `parseFloat` return `NaN`.


coryhouse:
Thanks for the PR! 👍

PR #260, opened 2016-09-15
Import path in webpack.config.dev.js

dtimberlake:
Need to import `path` to this file before using in the config.


coryhouse:
Good catch. Thanks! 👍

PR #267, opened 2016-09-30
Update test script to be more inclusive

kwelch:
Tests should now run all `.spec.js` files that are not in the `node_modules` folder.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,9 +20,9 @@
     "remove-dist": "rimraf ./dist",
     "prebuild": "npm run clean-dist && npm run lint && npm run test",
     "build": "babel-node tools/build.js && npm run open:dist",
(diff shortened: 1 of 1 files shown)


coryhouse:
Interesting. I'd never envisioned tests outside of src. What's the use case?

kwelch:
It came up from an odd scenario at our office, and I thought it may be useful. We had a different folder structure and there was a point of nesting that tests were inconsistency discovered. 

Our case was due to us not having out npm setup yet. We moved things some reusable code to a modules folder(sibling to src). Intent was cleaner distinction of the separation of application specific vs application agnostic. With the plan that it would be deleted as we moved the items into their own proper packages.

coryhouse:
Sorry I let this languish. Forgot to respond! Thanks Kyle!

kwelch:
No worries. 👍

PR #282, opened 2016-10-18
Use "babel-eslint" as parser for ESLint

BR0kEN-:
With Babel we can use a lot of syntax sugar - it's cool, but ESLint ain't able to fit the things.

For instance, class properties aren't available. More information can be found here: https://github.com/babel/babel-eslint/issues/312

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,6 +8,7 @@
   "plugins": [
     "react"
   ],
+  "parser": "babel-eslint",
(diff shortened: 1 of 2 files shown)


coryhouse:
I'm torn on this. I took this out because it's only necessary for those running experimental features. But I totally understand why some would want to use them, so I'll go ahead and accept. Thanks for the PR. 👍

Issue #284, opened 2016-10-23
setup script deletes .git repository


JohnLockwood:
Node version: 6.7.0 (or any)

npm version: 3.10.3 (or any)

Operating system: Ubuntu 16-ish (or any)

Command line used:  npm run setup

Steps to reproduce: Don't -- it'll delete your local git repository.  That's the point.

JohnLockwood:
Speaking of overzealous (Issue 18), the setup script tools/setup/setup.js contains the gem below, which I wish I'd known about before I forked the repo, cloned the fork, and spent a half hour working on a feature branch:

rimraf('.git', error => {
  if (error) throw new Error(error);
});
console.log(chalkSuccess('Original Git repository removed.\n'));

Fortunately it was only a half hour and I can get most of my changes back, and I do appreciate your effort in making this open source and your work on Pluralsight, but a bit of a heads up when you're getting ready to delete the git repo would be a nice touch.  People do store work there, and some of that work may not be pushed yet.

JohnLockwood:
C.f.:  https://github.com/CodeSolid/react-slingshot/commit/512be1c03077ce1b618abd319d7928c9f87871cf  -- for one possible solution.

coryhouse:
Ouch, I'm sorry about that John. Yes, I can see how that could surprise. I'd welcome a PR that adds a warning if you're willing. :)

JohnLockwood:
OK sent.  Not sure what's up with that node v4 test build yet though.

PR #285, opened 2016-10-24
Issue18

JohnLockwood:
Sorry, let me cancel this -- I need to cherry pick just the last commit.


JohnLockwood:
Sorry again for the confusion -- this is closed and shouldn't be merged.  But the commit "Change setup.js ..." might be one way of looking at Issue #284

PR #288, opened 2016-10-26
Nest prompts within git removal callback

hbaughman:
Given the existing structure, the "Original Git repository removed" message can (and on my machine *did*) appear after the prompt for the new project name. It's green styling arrested my attention from the prompt causing me to believe the process had hung. By moving the prompts inside the callback, git removal will always be completed before the prompting begins.

The confusing message is included below:

message screenshot

> Dependencies installed.
> WARNING:  Preparing to delete local git repository...
> prompt: Delete the git repository?  YES to continue or NO to skip.:  YES
> Updating package.json settings:
> prompt: Project name (default: new-project):  Original Git repository removed.


coryhouse:
Thanks! 👍

Issue #277, opened 2016-10-11
On a clean checkout of the repo, npm run test:cover doesn't create any files


rstudner:
```
git clone git@github.com:coryhouse/react-slingshot.git
cd react-slingshot
npm install
npm run test:cover
```

This leaves me with a coverage folder, with nothing inside of it.

Is this expected?

coryhouse:
Hrmm, that's unfortunate. Yes, I'm seeing this too, so something has regressed. I'm currently buried on a late project but would welcome a pull request from anyone with the time to debug this. 

Thanks for the heads up!

marinbcaro:
It looks like Isparta has already an open issue about this: https://github.com/douglasduteil/isparta/issues/126
I did what they suggested: adding "istanbul": "0.4.4" to package.json and that worked for me. @coryhouse should I create a pull request for this or wait until the issue is fixed by Isparta?

coryhouse:
Fixed in #294. Thanks @marinbcaro!

PR #294, opened 2016-10-31
Fix to make npm run test: cover create files

marinbcaro:
pull request for #277


coryhouse:
Wow, simple as that?! Thanks for the PR! 👍

Issue #259, opened 2016-09-14
Debugging with VSCode


cristian-eriomenco:
First of all I want to give many respects to the authors and contributors, the setup is simply amazing and what a real world requires.

Taking a step further I know we can use chrome dev tools for debugging, but IMHO would be very nice to have the debugger setup to be working on VSCode (although people might be using other editor). 

There are already some attempts out there:
1. https://github.com/skolmer/react-hot-boilerplate-vscode
2. https://github.com/Microsoft/vscode-chrome-debug
3. http://stackoverflow.com/questions/31711286/vscode-debug-es6-application
4. https://gist.github.com/dchowitz/83bdd807b5fa016775f98065b381ca4e

Still they feel tied to the specifics of their own setup. 
I've tried to get it working but seems that my knowlegde is not good enough. 

So if this request seems sensible for the community please upvote it, or at least express your thoughts upon.

Thanks!

nickytonline:
@cristian-eriomenco, have you tried following the steps outline here? https://code.visualstudio.com/blogs/2016/02/23/introducing-chrome-debugger-for-vs-code

nickytonline:
@cristian-eriomenco, are you still having issues with debugging in VS Code?

kwelch:
If there were any changes made that were needed to get it working, it would be awesome to see a PR that add it is or adds documentation on the steps necessary.

nickytonline:
@kwelch, I'll take a look at this this weekend.

nickytonline:
@cristian-eriomenco, I can confirm that with the Chrome Debugger for VS Code which I suggested using above, you cannot  debug react-slingshot in VS Code. Just investigating why.

nickytonline:
@cristian-eriomenco, cancel that. It does work. I had the wrong port configured. Here's is a sample launch.json file for debugging in VS Code.

```
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Launch Chrome against localhost, with sourcemaps",
      "type": "chrome",
      "request": "launch",
      "url": "http://localhost:3000",
      "sourceMaps": true,
      "webRoot": "${workspaceRoot}"
    },
    {
      "name": "Attach to Chrome, with sourcemaps",
      "type": "chrome",
      "request": "attach",
      "port": 9222,
      "sourceMaps": true,
      "webRoot": "${workspaceRoot}"
    }
  ]
}
```

[...]

PR #300, opened 2016-11-06
Added blurb about how to debug in VS Code.

nickytonline:
@coryhouse, @kwelch, I've updated the README.md to explain the steps on how to debug in VS Code. Just give it a quick read to let me know if it sounds OK.


nickytonline:
This fixes #259.

coryhouse:
Great. Suggest moving this blurb to the FAQ instead of the main page. Also suggest dropping the last bullet about issues for brevity. (Goal is to have a clean front page with only the critical info).

I also want to organize the FAQ with anchors at the top since it's gotten long, but that's a separate to do for another day.

nickytonline:
OK, thanks for the feedback. I'll make the change.

nickytonline:
@coryhouse, I've moved it to the FAQ.md with your suggestions.

coryhouse:
Thanks Nick!

Issue #299, opened 2016-11-05
Update Redux DevTools Extension to support latest version


kwelch:
Redux DevTools extension is deprecating the method that slingshot is using to integrate. 

[Here is the example on setup similar to how we already.](https://github.com/zalmoxisus/redux-devtools-extension#12-advanced-store-setup)

~~Ideally, we should support both versions.~~

zalmoxisus:
Hey @kwelch,

We just renamed `window.devToolsExtension` to `window.__REDUX_DEVTOOLS_EXTENSION__` as there were complains as it being to generic.

Also added a second method  `window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__` which is more suitable for your case. It makes the integration with other enhancers and middlewares better, specifically the ability to lock the app (anhancers and middlewares) when the `lock` button is pressed, and for remotely dispatched action creators to be passed via enhancers/middlewares. Also there's no need for the noop function (`f => f`).

I think supporting the deprecated global variable isn't necessary as Chrome extension is updated automatically. Only for Firefox is possible to disable the automatic updates, but [it's been more than a month since we made those changes](https://github.com/zalmoxisus/redux-devtools-extension/issues/220).

kwelch:
@zalmoxisus, 

The upgrade was very simple, I just didn't have a chance to do it for slingshot yet. I agree, with autoupdates and it has been deployed for over a month it makes sense to not support both. 

The extension is extremely useful. Thanks for you work on it and your added details here.

nickytonline:
@kwelch, PR is up for this.

PR #305, opened 2016-11-09
Upgrade for the Redux DevTools extension.

nickytonline:
A little OSS during my coffee break. Implements #299.

### DIFF
--- a/src/store/configureStore.dev.js
+++ b/src/store/configureStore.dev.js
@@ -19,9 +19,9 @@ export default function configureStore(initialState) {
     thunkMiddleware,
   ];
 
-  const store = createStore(rootReducer, initialState, compose(
-    applyMiddleware(...middlewares),
-    window.devToolsExtension ? window.devToolsExtension() : f => f // add support for Redux dev tools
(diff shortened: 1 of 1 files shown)


src/store/configureStore.dev.js
```diff
@@ -19,9 +19,9 @@ export default function configureStore(initialState) {
     thunkMiddleware,
   ];
 
-  const store = createStore(rootReducer, initialState, compose(
-    applyMiddleware(...middlewares),
-    window.devToolsExtension ? window.devToolsExtension() : f => f // add support for Redux dev tools
+  const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose; // add support for Redux dev tools
```
kwelch: Do you think we should but a sanity check around `window` being defined?

nickytonline: I think it's safe to assume `window` will be defined and even if you hit this code while running tests, `window` can be mocked in your tests.

zalmoxisus: Definitely, if it's not an universal (isomorphic) app.

PR #313, opened 2016-11-12
making configureStore compatible with Webpack 2

mlwigdahl:
This is just the change to configureStore to remove the dynamic inclusion.  I'll bundle the rest of the Webpack 2.0 changes in a separate PR later today.


src/store/configureStore.js
```diff
@@ -1,5 +1,58 @@
...
+      store.replaceReducer(nextReducer);
+    });
+  }
+
+  return store;
+}
+
+let configureStore = function () {};
+
 if (process.env.NODE_ENV === 'production') {
```
coryhouse: Suggest simplifying via ternary:

const configureStore = process.env.NODE_ENV === 'production' ? configureStoreProd : configureStoreDev;

coryhouse:
Made a minor suggestion to streamline the code. Other than that, looks great. Thanks! 👍

coryhouse:
Thanks so much! 👍

dvanoni:
By combining the dev and prod configurations into a single file, isn't this re-raising the concern that @coryhouse brought up in #182?

kwelch:
I think it is valid to consider dev be bundled in a prod build, but I believe it is moot in this case since it was done for Webpack 2 which has tree-shaking which will eliminate the unused dev code.

mlwigdahl:
Yes, that would be a consideration.  I don't actually know whether Webpack 2 will properly shake the dev dependency out or not -- I suspect not, actually, since the function is referenced in code that can't be statically analyzed for unreachability.  But Webpack 2 can't do tree shaking unless you turn off CommonJS transpilation of ES6 import statements, which breaks the two-file approach.

It's kind of a Catch-22 situation, and we may very well need a better approach here to get the best of both worlds.

coryhouse:
This is an interesting conversation and one that will be common as tree shaking matures. I agree with @kwelch that tree shaking in Webpack 2 should compile this out. It's easy to confirm, but about to jump on the flight to Nodevember!

PR #314, opened 2016-11-12
Webpack 2

mlwigdahl:
OK, this is the sum total of what I did to support Webpack 2.  This subsumes the PR I submitted earlier that just included the configureStore changes.


coryhouse:
Awesome! Thanks so much for sharing! Very tempted to merge this. Must...wait...for...release :) 👍

coryhouse:
Wow, I'm surprised at the subtle difference I'm seeing with Webpack 2:
241k Webpack 1
238k Webpack 2

I expected tree shaking to have a bigger impact. Perhaps this is because the example app is using pretty much all of React and the other libs we're referencing. But this doesn't smell right...

mlwigdahl:
I know what you mean.  I experimented a bit by importing lodash in one of the source files and using _.head on an array.  Webpack 1's bundle increased to 305 kB and Webpack 2's bundle was 302 kB.  The head function itself is maybe 50 bytes and stands alone, so clearly I don't have the Webpack 2 tree shaking working properly.

I'll continue working on it and let you know when I get it working.

mlwigdahl:
Did find a reference that indicates we may not get good results from tree-shaking until other libraries are updated to use it:

> Tree shaking will only work if you're using ES6 import/exports, a la:
> 
> import { somethingSpecific } from 'some/package';
> If you're using CommonJS (e.g. require()) or pulling everything in, like import * as whatever from 'wherever', then you won't get the benefit of tree-shaking-- you'll be pulling in everything that the module exports.
> 
> Tree shaking will start to make a lot of sense when third-party libraries adopt ES6 en masse. Some have started offering ES6 and CommonJS versions (usually by importing from a different file, or with their package.json file having a different entry point); use them where possible.

[https://github.com/webpack/webpack/issues/3092]

ryannealmes:
Seems like webpack 2 has been released! It would be awesome to see the PR merged :)

kwelch:
I believe we will want to wait until the week grace period has ended and it is released under the latest tag. Additionally there are conflicts that must be addressed.

iLeonelPerea:
Ready to merge :D

kwelch:
@mlwigdahl Can you get conflicts resolved on this?

@coryhouse Are you comfortable with this?
I think this and Jest would be a good point to release a v6.

coryhouse:
I agree, this would be a great combo for a major release! 🎉

kwelch:
Another good add would be react-router v4, but I am not sure when that is coming out of beta. 

I have really enjoyed using it.

coryhouse:
Agreed, but yeah, gotta wait for now.

kwelch:
Conflicts are resolved, but this deserves just another pass of testing.

mlwigdahl:
Great, kwelch, is there anything I need to do at this point?  Are the CI failures still an issue?

kwelch:
I may not have merged properly, I did it in the browser. If you could pull down and confirm functionality, that way I can merge then integrate with the #363 - Jest Migration.

kwelch:
@mlwigdahl I attempted to pull this down and ensure everything was working, but I was getting a few errors. I have not done anything with `webpack@2` yet, so I don't feel comfortable with the changes necessary. I do know a simple one is that you can no longer emit `-loader`, but I was getting errors in lint and tests as well.

webpack.config.dev.js
```diff
@@ -36,7 +34,19 @@ export default {
...
+        sassLoader: {
+          includePaths: [path.resolve(__dirname, 'src', 'scss')]
+        },
+        context: '/',
+        postcss: () => [autoprefixer],
+      }
+    }) 
   ],
   module: {
     loaders: [
```
ollwenjones: Per [docs](https://webpack.js.org/guides/migrating/) `module.loaders` is now `module.rules`

webpack.config.dev.js
```diff
@@ -50,6 +60,5 @@ export default {
       {test: /(\.css|\.scss)$/, loaders: ['style', 'css?sourceMap', 'postcss', 'sass?sourceMap']},
```
ollwenjones: Heard of others having trouble with webpack 2 and the loader shorthand (sans `-loader` suffix) and having to add the `-loader` suffix everywhere. Possible that's been resolved since then, though.

webpack.config.prod.js
```diff
@@ -64,7 +59,20 @@ export default {
...
+        sassLoader: {
+          includePaths: [path.resolve(__dirname, 'src', 'scss')]
+        },
+        context: '/',
+        postcss: () => [autoprefixer],
+      }
+    })
   ],
   module: {
     loaders: [
```
ollwenjones: same comment about `module.rules` 👆

webpack.config.dev.js
kwelch: There is a way to support emitting it, but you need to use `resolveLoader.moduleExtensions`. See https://webpack.js.org/guides/migrating/#automatic-loader-module-name-extension-removed

ollwenjones:
Looks like DedupePlugin [should be removed](https://webpack.js.org/guides/migrating/#dedupeplugin-has-been-removed)

webpack.config.prod.js
```diff
@@ -64,7 +59,20 @@ export default {
     new webpack.optimize.DedupePlugin(),
 
     // Minify JS
-    new webpack.optimize.UglifyJsPlugin()
+    new webpack.optimize.UglifyJsPlugin(),
```
ollwenjones: I also had to add an options object to the UglifyJsPlugin constructor for the build to succeed.
```
    new webpack.optimize.UglifyJsPlugin({
      sourceMap: true
    }),
```
Per [the docs](https://webpack.js.org/guides/migrating/#uglifyjsplugin-sourcemap)

webpack.config.dev.js
```diff
@@ -50,6 +60,5 @@ export default {
...
       {test: /\.json$/, loader: "json"}
```
ollwenjones: https://webpack.js.org/guides/migrating/#json-loader-is-not-required-anymore

mlwigdahl:
Sorry for the delay, wasn't able to get to this earlier.  I fixed up a couple of issues that appear to have come in on the merge, as well as redoing the webpack config files to match the requirements of the release version of Webpack2.  Assuming the tests pass, we should be good to go.

coryhouse:
@mlwigdahl It doesn't look like your code reflects the changes suggested by @ollwenjones yet.

mlwigdahl:
Yep, looks good!  Let me know if we need to address anything else.

mlwigdahl:
Must have missed those.  Let me check.

mlwigdahl:
Ah, I missed adding the sourceMap option to UglifyJsPlugin.  I also didn't change "loaders" to "rules", as loader syntax is still supported in this build, but I can do that if you want.  Let me get my kids to bed, then I'll get these changes in.  :)

mlwigdahl:
OK, added the changes @ollwenjones had suggested that weren't already included, plus fixed a typo.  It would probably be possible to improve the loader setup a bit more -- Webpack2 supports a less "string-y" configuration setup, but I'm not confident enough in my knowledge of it to pull the trigger at the moment.  Other than that, I think this is everything.

kwelch:
I will pull this down and look at in later tonight or tomorrow. I have a few other things to wrap up.

coryhouse:
Thanks for all your work on this @mlwigdahl! I just merged @kwelch change to Jest, so those changes will need to be integrated into this branch before this is pulled in.

mlwigdahl:
Just merged with the Jest code, no major conflicts.

kwelch:
This looks good to me. 👍

.babelrc
```diff
@@ -1,20 +1,33 @@
...
   ],
   "env": {
     "development": {
       "presets": [
+        "latest",
         "react-hmre"
       ]
     },
     "production": {
+      "presets": [
```
coryhouse: Why is this change now necessary? Is this change related to the move to Webpack 2?

mlwigdahl: Yes, since the production build needs the default CommonJS module support from the ES2015 plugin that is included in the "latest" plugin disabled for tree-shaking, I needed to:

1. Distribute the "latest" preset to all three different levels and 
2. Override the module support setting for production only.

The result is that production builds use untransformed ES6 module load syntax, which Webpack can tree-shake.  The other builds use CommonJS transformation as they did before.

coryhouse: Ah, makes sense. Thanks for the detailed explanation!

ollwenjones: @mlwigdahl a very key, but not obvious catch. Good job.

PR #316, opened 2016-11-14
Make removeDemo.js remove itself

toddprater:
Currently, the removeDemo.js script removes demo app code and other
artifacts, but does not delete itself nor the 'remove-demo' script
entry in package.json. This commit takes care of that last bit of
clean up.

### DIFF
--- a/tools/removeDemo.js
+++ b/tools/removeDemo.js
@@ -16,7 +16,8 @@ const pathsToRemove = [
   './src/store/store.spec.js',
   './src/styles',
   './src/routes.js',
-  './src/index.js'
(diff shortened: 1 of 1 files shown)


toddprater:
Oops. I see I'm going to have to figure out this new 'yarn' thing-a-ma-jiggy. :) Will see if I can get this to build on AppVeyor tomorrow...

coryhouse:
It appears the failure in Appveyor was bogus (likely due to a sporadic network issue), so I've accepted as is. Great call on this change! 👍 

Thanks so much! :)

PR #317, opened 2016-11-14
Fix lint warning introduced by #316

toddprater:
There were a couple of eslint warnings related to 'var' usage in the previous commit for #316. (Sorry! Old habits die hard. Probably would have shown up in the appveyor output if it had not glitched out? Or if I had paid more attention on my own local build.)


coryhouse:
Happens to the best of us! Thanks for the quick PR! :)

Issue #323, opened 2016-11-19
Update Travis CI configuration


nickytonline:
@coryhouse, Travis CI just merged this, https://github.com/travis-ci/travis-build/pull/895 and it's rolling out Monday, so we should update the Travis CI configuration sometime after Monday since yarn will now be preinstalled.

coryhouse:
Good catch! 👍

kwelch:
Closing as PR was merged.

PR #327, opened 2016-11-22
updates to travis ci config for yarn.

nickytonline:
Not ready to merge yet, just testing if CI runs properly. This implements enhancement #323

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,9 @@
 language: node_js
-before_install:
-  # Repo for Yarn
-  - sudo apt-key adv --keyserver pgp.mit.edu --recv D101F7899D41F3C3
-  - echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
-  - sudo apt-get update -qq
-  - sudo apt-get install -y -qq yarn
-install:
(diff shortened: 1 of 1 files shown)


nickytonline:
The change to the configuration works, but since we don't add an install directive, the default is run which runs

`yarn` and `npm test` which is not what we want because we have an after success step that runs coveralls and runs the tests. As they state on their [blog post](https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn), "Please note that, at this time, Yarn is not yet pre-installed on our build images. This means that, if yarn.lock does not exist in the expected location, you are responsible for installing Yarn yourself."

Let's wait to merge this once it's pre-installed so we can just run 'yarn' at the install phase.

coryhouse:
Cool, and agreed. 👍

cyprusglobe:
@nickytonline @coryhouse might take a look now looks like its preinstalled for trusty according to this https://docs.travis-ci.com/user/build-environment-updates/2016-12-01/

nickytonline:
Thanks @cyprusglobe. I'll test it out.

nickytonline:
@coryhouse, this is good to merge now.

coryhouse:
Thanks Nick! 👍

PR #329, opened 2016-11-23
removed extra -- on lint/test:watch

Vlandmist:
After clean install and project setup (on Node 7.2) I was getting errors on npm start. (ERROR: "lint:watch" exited with 1.)

Removing the extra -- solved it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "open:src": "babel-node tools/srcServer.js",
     "open:dist": "babel-node tools/distServer.js",
     "lint": "esw webpack.config.* src tools --color",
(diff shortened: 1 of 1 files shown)


coryhouse:
The extra -- is necessary to pass the watch flag to lint and test. This way they run each time the user hits save. If you clone the project from scratch, it should work fine. If it doesn't, please open an issue and we'll work with you there.

Vlandmist:
Ok sorry for that. I had cloned from scratch (installed everything a few hours ago) but I'll try again and open an issue in case it happens again.

PR #336, opened 2016-12-06
Add JSDOM and config in testSetup to allow the use of Enzyme's mount …

silentbobbert:
…function in addition to the shallow function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "isparta": "4.0.0",
     "istanbul": "0.4.4",
     "json-loader": "0.5.4",
+	"jsdom": "8.5.0",
     "mocha": "3.1.2",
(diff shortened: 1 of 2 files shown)


kwelch:
I think this is a great addition. Would you be able to add a few tests that show the how/why to use this?

silentbobbert:
I've been watching Cory's pluralsight courses and just learning React. I have been refactoring the demo app to split mark up out into presentational containers. That lead me to add JSDOM to test parent/child containers. I'm not sure if if Corey et al would want to take this change or not. I'm having trouble at the moment passing an errors object down on props so I don't have a working example to submit yet.

coryhouse:
Looks great! Thanks! 👍

PR #337, opened 2016-12-07
Updated badges

jasonnathan:
Build status was not accurate and updated all to use shields.io


coryhouse:
Thanks Jason! 👍

kwelch:
Question: Is it possible to have the badges dynamic based on the branch the user is looking at? Just wondering since this is hardcoded to master.

jasonnathan:
Sure, you'd just have to update the Readme.md per branch. Pull request? :P

kwelch:
I don't think that is necessary, wasn't sure if there was a variable that could be used in place of master to do it "auto-magically"

PR #338, opened 2016-12-08
Replace var with const in testSetup script

fayimora:
Got this after running `yarn start`.


coryhouse:
Good catch. Thanks! 👍

PR #346, opened 2016-12-17
Update eslint import rules

fayimora:
See 

* https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-extension.md#restrict-file-extensions-that-may-be-required-require-extension
* https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-wrap-multilines.md


coryhouse:
Thanks! 👍

PR #348, opened 2016-12-18
[WIP] Change dist server to run on port 4000

fayimora:
I think it makes sense to want to build while dev server is running(I do this periodically). Unfortunately, the dist and dev server run on the same port so this fails. This PR changes the port the dist server runs on so both can run at the same time.

Questions
==

* Is this a desired change? I'm not sure if there was a reason for having them on the same port in the first place
* Is port `4000` ok? I'm happy to change to something else.

Issues
==
Executing `yarn start` or `npm start` will remove the dist folder. This is not good if the dist server is running and its files are getting deleted. My solution is to not execute `remove-dist` when the dev server is started but that depends on why it currently happens.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "setup": "node tools/setup/setupMessage.js && npm install && node tools/setup/setup.js",
     "remove-demo": "babel-node tools/removeDemo.js",
     "start-message": "babel-node tools/startMessage.js",
-    "prestart": "npm-run-all --parallel start-message remove-dist",
+    "prestart": "npm-run-all --parallel start-message",
     "start": "npm-run-all --parallel test:watch open:src lint:watch",
     "open:src": "babel-node tools/srcServer.js",
     "open:dist": "babel-node tools/distServer.js",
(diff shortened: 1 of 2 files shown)


coryhouse:
Hi @fayimora. I see your point. Great idea. The only reason dist is being deleted is to remind people to run dist before committing. I'd be okay with eliminating dist removal during npm start. 

@nickytonline @kwelch thoughts?

nickytonline:
@fayimora @coryhouse changing the port seems fine and we can remove the deletion of the `dist` folder, but I don't generally run into this scenario since my CI would generate the build artifacts (`dist`).  Having said that, I don't see it harming anything since CI will still work (fresh install each time). @kwelch?

kwelch:
I am good with this all around. 👍

coryhouse:
Great. @fayimora if you submit an updated PR that eliminates the remove-dist step I'll get it merged. Thanks!

fayimora:
@coryhouse done!

coryhouse:
Looks great! Thanks so much @fayimora! 👍

fayimora:
My pleasure! 🙇

PR #355, opened 2017-01-11
Fix appveyor cahce path

kwelch:
When I was playing with the jest-migration I noticed an error that path was not defined. It does not stop the build from being successful and I did not see the error in any other builds. I still couldn't leave it untouched.


coryhouse:
Thanks! 👍

PR #358, opened 2017-01-25
Bugfix for test coverage

mherodev:
Changed the target version for istanbul and updated the configuration to cover all the files, retargeted to `mocha` instead of the `node_modules` mocha path, and replaced the regex syntax with a cleaner nested function call (edit: correction, that breaks Windows builds, so it's a simple path now). 

This was only tested on OSX, so I don't know if the latter introduces Windows issues.

And apologies if I did anything wrong, this is my first open source PR 😨 

Edit: How do I tie this PR to the issues it addresses? Do we want commits squashed, too?


coryhouse:
Thanks so much! 👍

silentbobbert:
I just tried in on a Window 2012 build server and it doesnt work there either :(

Perhaps we need to help with the move to Jest if its the better, well-rounded framework?

Issue #361, opened 2017-02-03
Update Jest Reporter


kwelch:
The default jest reporter is "loud" and makes it possible to miss lint errors when using commands like `start` or `build`. 

Jest currently only allows for the standard reporter. Once they allow for [other reporters](https://github.com/facebook/jest/pull/2703#issuecomment-275475570), Slingshot will need to be updated to utilize it.

nickytonline:
Once https://github.com/facebook/jest/pull/3064 is merged, we can do this up.

coryhouse:
Great to see it's on it's way! We'd benefit most from a single line reporter, so hopefully that's created soon.

abdulhannanali:
@nickytonline Thanks for mentioning this! It's on it's way. Good to see it's going to be beneficial.

nickytonline:
@kwelch, are we moving forward with this still?

coryhouse:
Nice timing Nick - I'm working on this on a work project, so I can port the change over here when I'm done. :)

nickytonline:
Cool beans. 💯

PR #362, opened 2017-02-03
Replace Mocha with Jest

kwelch:
Migration complete. 

Only item left will is switching to a new reporter which I have opened #361, for tracking purposes.


kwelch:
I messed this up. Closing.

PR #363, opened 2017-02-03
Jest migration

kwelch:
Migration complete.

Only item left will is switching to a new reporter which I have opened #361, for tracking purposes.


package.json
```diff
@@ -44,36 +44,31 @@
     "babel-cli": "6.18.0",
     "babel-core": "6.20.0",
     "babel-eslint": "7.1.1",
+    "babel-jest": "^18.0.0",
```
coryhouse: A few of these have carrots. I've leaned toward specifying exact version for consistency. But I could be swayed. Either way, we should handle all these the same.

silentbobbert: I agree with Cory - some packages dont follow versioning properly and introduce breaking changes when they shouldnt...it is best to be specific and control the version you upgrade to next

nickytonline: This is why I put `.npmrc` with `save-exact=true` in place near the beginning of the project. However,  since introducing yarn, it ignores `.npmrc`.  Does anyone know the equivalent in yarn?

nickytonline: There appears to be an open issue about this for Yarn, https://github.com/yarnpkg/yarn/issues/2193

package.json
```diff
@@ -20,11 +20,11 @@
     "remove-dist": "rimraf ./dist",
     "prebuild": "npm run clean-dist && npm run lint && npm run test",
     "build": "babel-node tools/build.js && npm run open:dist",
-    "test": "mocha tools/testSetup.js src/**/*.spec.js --reporter progress",
-    "test:cover": "babel-node node_modules/isparta/bin/isparta cover --root src --report html mocha -- --require ./tools/testSetup.js src/**/*.spec.js --reporter progress",
-    "test:cover:travis": "babel-node node_modules/isparta/bin/isparta cover --root src --report lcovonly _mocha -- --require ./tools/testSetup.js src/**/*.spec.js && cat ./coverage/lcov.info | node_modules/coveralls/bin/coveralls.js",
+    "test": "node --harmony_proxies node_modules/jest/bin/jest",
+    "test:cover": "npm run test -- --coverage ",
+    "test:cover:travis": "npm run test:cover && cat ./coverage/lcov.info | node_modules/coveralls/bin/coveralls.js",
```
kwelch: @coryhouse I assume you added this, can you elaborate on what it is and confirm if it is needed.

src/utils/__snapshots__/fuelSavingsCalculator.spec.js.snap
```diff
@@ -0,0 +1 @@
+exports[`Fuel Savings Calculator milesPerMonth throws an error on invalid timeFrame 1`] = `[Error: Unknown milesDrivenTimeframe passed: minute]`;
```
kwelch: This was suppose to fail. Updating it from a snapshot to a `.toThrow`

package.json
coryhouse: I pulled that from here: https://github.com/nickmerwin/node-coveralls/issues/126

The coveralls docs lack info on Jest integration, but I found this example in that issue.

kwelch: I removed that and the lcov.info is still being produced. Will confirm the coverage once checks are completed.

kwelch:
The `lcov.info` is still used, removing causes it fail.

kwelch:
@coryhouse Is there anything else left that you see needs to be addressed for this?

coryhouse:
Merged! Thanks so much to @kwelch for all the hard work on this! 🥇

kwelch:
Thanks! I am excited about Jest. Giving a talk at a meetup later this month. 😄

coryhouse:
We're using Jest at work and I'm using it in my upcoming course. Been really happy with it so far. Though I am excited to get the minimal output support added.

kwelch:
I am hoping to wrap a few things up and focus directly on the effort. 👍

silentbobbert:
Hey @kwelch I noticed you removed the test setup file. I previously contributed a section of that file which set up jsdom so that you could use mount as well as shallow from enzyme. Is Jest setting that up implicitly now?

kwelch:
Yes, Jest's default environment is JSDOM. From the work I have done with it, it appears that JSDOM is not directly configurable through the Jest config. 

Did you happen to put specific tests in that exampled out why the mount was needed? If so, all component test are still passing as expected.

silentbobbert:
Thanks @kwelch . My contribution did not contain any new or modified tests - I back ported it from a project I started using slingshot as a template, and thought it would be useful for others who did the same. There was no actual need for mount (or jsdom) in the fuel savings example app, but for people using slingshot as a template, they would quickly need it for a more serious app, hence why I thought it was a useful contribution none the less.

kwelch:
I agree, it is helpful. Would you be willing to PR with a add test that leverages mount? This will help confirm the functionality and can also be used as an example on how to use it for those new to component testing. 😃

Issue #366, opened 2017-02-12
node: bad option: --harmony_proxies


arshdkhn1:
Node version: v7.5.0

npm version: 4.1.2

Operating system: Elementary OS Luna

Command line used: Hyper / 

Steps to reproduce: `npm start -s`

Hi, I get `node: bad option: --harmony_proxies` as error when I try to start the app. When I downgraded my node to lts 6.9.5 then the error was gone.

Creating this issue to bring it to your attention if it hasn't yet.

XOP:
Same for me with `7.2.0`

dotnetnorm:
Remove it from test wit 7.2.0 it is not needed.  It is on the test mpmaria@us.ibm.com script.

n-makarov:
Node 7 has removed the `harmony_proxies` flag in favor of `harmony`. You need to update `test`  command in the `package.json` file by replacing `--harmony_proxies` with `--harmony`.

kwelch:
This flag is not necessary for v6 of node either, but was replaced with a no-op. It was deprecated in 7 as @n-makarov stated. 

Due to slingshot supporting Node v4 and v5, this is a necessary flag.

I do understand the pain point, if someone would like to update docs around this that a PR would be greatly appreciated.

coryhouse:
I believe I've resolved the issue for local dev via #368. 

Still looking for a fix for CI build on Node 7. Open to ideas.

nickytonline:
@coryhouse, just typing out loud here ;). We could have a script that detects which version of node is running via `const requiresHarmonyFlag = parseInt(/^v(\d+)\./.exec(process.version)[1], 10) >= 7` and if it's true, we run `node node_modules/jest/bin/jest`, otherwise we run `node --harmony_proxies node_modules/jest/bin/jest`.

coryhouse:
Yeah, that'd do it. But man, it'd be ugly and confusing to put all that in an npm script. Would be more acceptable abstracted into a js file...

nickytonline:
@coryhouse, yeah, I meant in a JS file. Forgot to finish typing my thoughts ;)

nickytonline:
I'll throw up a PR for this this weekend.

nickytonline:
I thought I'd get to this this weekend but did not. I'll try and knock this off this week.

coryhouse:
Excellent 🎉  - Eager to turn 7 back on for CI. :)

coryhouse:
[...]

PR #367, opened 2017-02-13
updated readme about node v7 harmony_proxies flag issue

arshdkhn1:
#366

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ This will run the automated build process, start up a webserver, and open the ap
 7. **Having issues?** See "Having Issues?" below.
 
 ##Initial Machine Setup
-1. **Install [Node 4.0.0 or greater](https://nodejs.org)** - (5.0 or greater is recommended for optimal build performance). Need to run multiple versions of Node? Use [nvm](https://github.com/creationix/nvm).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ This will run the automated build process, start up a webserver, and open the ap
 7. **Having issues?** See "Having Issues?" below.
 
 ##Initial Machine Setup
-1. **Install [Node 4.0.0 or greater](https://nodejs.org)** - (5.0 or greater is recommended for optimal build performance). Need to run multiple versions of Node? Use [nvm](https://github.com/creationix/nvm).
+1. **Install [Node 4.0.0 or greater upto Node 6.x.x](https://nodejs.org)** - (5.0 or greater upto 6.x is recommended for optimal build performance). Node 7 isn't supported yet because it has removed the harmony_proxies flag in favor of harmony. `--harmony_proxies` is a necessary flag for Node 3 and Node 4. Need to run multiple versions of Node? Use [nvm](https://github.com/creationix/nvm).
```
kwelch: May want to include Node 5. Travis builds failed without this for that version as well.
https://travis-ci.org/coryhouse/react-slingshot/builds/197890657

kwelch:
@coryhouse Do you agree to say that it is not supported? If so, I believe we should also update the `nodeVersionCheck.js`

coryhouse:
I'd suggest just mentioning changing the test script if using Node 7. It's otherwise supported, right?

kwelch:
Agreed. I think the line in the PR makes it sound completely unsupported where only that flag within that script is unsupported. 

It may make sense to put a warning in the nodeVersionCheck and not an error to help inform users of the unsupported flag.

coryhouse:
Closing since handled via #368

PR #368, opened 2017-02-14
Fix Jest issue on Node 7.

coryhouse:
Eliminating the harmony-proxies reference on the plain test script since (I think) it's only needed for Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: node_js
 cache: yarn
 node_js:
+  - "7"
   - "6"
   - "5"
   - "4"
-after_success:
-  # Send coverage data to coveralls.
-  - yarn run test:cover:travis
+# Run tests and send coverage data to coveralls.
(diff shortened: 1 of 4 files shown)


kwelch:
You will probably need to update the `.travis.yml`, it is currently using the default of `npm test`

Also, should we be testing that it builds? Current script is not doing this.

coryhouse:
Good point. How'd the never occur to me?!

kwelch:
Don't forget about appveyour. Also may want to have a custom `build:ci`, that uses `test:travis` and does not open the dist-server.

coryhouse:
Merging since this fixes the Node 7 issue for those running locally. Node 7 on Travis and Appveyor are failing [due to this issue](https://github.com/adonisjs/adonis-framework/issues/357), but that only effects CI. Will figure out how to fix the CI in a separate ticket.

PR #372, opened 2017-02-17
Adding sass indented syntax support.

arshdkhn1:
I hope these are the only files that I have to change in order to support indented sass syntax. Please point out if I am missing anything.

### DIFF
--- a/src/components/HomePage.js
+++ b/src/components/HomePage.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import {Link} from 'react-router';
+import '../styles/homepage.sass';
 
 const HomePage = () => {
   return (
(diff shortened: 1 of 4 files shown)


src/components/HomePage.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import {Link} from 'react-router';
+import '../styles/homepage.sass';
```
arshdkhn1: A demo file to show sass indented syntax support.

coryhouse:
Great! Can you please remove the actual Sass example though? I figure merely supporting it via the loader is sufficient.

arshdkhn1:
I am not sure what this code does - 
``` js
sassLoader: {
          includePaths: [path.resolve(__dirname, 'src', 'scss')]
        },
```
I don't see any such folder. What does this code do?

arshdkhn1:
@coryhouse removed the sass example.

coryhouse:
Squashed and merged. Thanks!

PR #385, opened 2017-02-27
Bumped sass-loader to v6.0.2

ryannealmes:
Since sass-loader was sitting on an old version (4.0.0) I thought it prudent to update it since I ran into a bunch of issues with the older version.


kwelch:
You could switch to the master branch, pull down latest and rebase this branch using master. 

We have updated to v6.0.0, and it appears your version is still using v5.0.0

Thanks for the update 👍

ryannealmes:
I will create a new branch and sort this out tomorrow.

PR #386, opened 2017-02-28
Bumped the sass loader version

ryannealmes:
Since sass-loader was sitting on an old version (4.0.0) I thought it prudent to update it since I ran into a bunch of issues with the older version.


coryhouse:
Thanks! 👍

PR #390, opened 2017-03-07
Updated `extract-text-webpack-plugin` to latest version

jkinzel-r7:
The beta version had an issue where the options passed were not applied properly. See this for more info: https://github.com/webpack-contrib/sass-loader/issues/388 

This doesn't impact the demo, but if folks tweak the configuration it may trip them up. It was also the last remaining beta-stage plugin in the project.


coryhouse:
Thanks! Looks like Yarn's server is down which broke the CI build, but accepting. Thanks! 👍

PR #216, opened 2016-07-28
Better HMR with React Hot Loader v3

paulwithap:
Added React Hot Loader v3 beta to the project for hot reloading components.

PR #392, opened 2017-03-14
Add React Hot Loader 3

oshalygin:
- This commit wires up react-hot-loader 3 to
  be used in the application.  There are numerous
  benefits to the latest release, all of which
  can be seen at https://github.com/gaearon/react-hot-loader
- Note that the specific implementation around
  wrapping in a Root component is part of how
  react-hot-loader 3 needs to be configured.
- Note that the package is brought in as a
  dependency, not a dev dependency because of
  how it is switched at runtime or not.
- More information on the migration can be
  viewed at:
https://github.com/gaearon/react-hot-loader/tree/master/docs

Related: #216

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "postcss-loader": "1.2.1",
     "prompt": "1.0.0",
     "react-addons-test-utils": "15.4.1",
+    "react-hot-loader": "3.0.0",
     "redux-immutable-state-invariant": "1.2.4",
     "replace": "0.3.0",
     "rimraf": "2.5.4",
@@ -86,7 +87,7 @@
     "webpack": "2.2.1",
     "webpack-bundle-analyzer": "2.1.1",
     "webpack-dev-middleware": "1.9.0",
-    "webpack-hot-middleware": "2.13.2",
+    "webpack-hot-middleware": "2.17.1",
     "webpack-md5-hash": "0.0.5"
   },
   "keywords": [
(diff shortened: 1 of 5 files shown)


coryhouse:
Excellent! Thanks for the PR! This looks great to me but I'm going to let it sit for a couple days so hopefully others can chime in. Thanks again!

oshalygin:
Sounds good thanks @coryhouse !

kwelch:
I put in my comment on the version number, other than that all appears to be accurate.

oshalygin:
@kwelch watcha mean by version number?  I didn't see your comment :)

kwelch:
In the `package.json` instead of having a specific version number mentioned it just says `"next"`. 

It is an oddity with `yarn`, I put in a ticket yarnpkg/yarn/#1306, but it appears to have resurfaced. Additionally, yarn does not use the `.npmrc` so it does not save exact and as the point that this was last discussed in #363, they had not set on a solution.

oshalygin:
I'll update it, thanks for the tip!

oshalygin:
thanks @kwelch

kwelch:
Thanks, we appreciate the PR. Although, I have used v3 while it was in beta and had no issues. I am with Cory on leaving this open for a little while to ensure it gets proper eyes on it.

kwelch:
Is there anything holding this PR up?

coryhouse:
Ah, forgot to merge this after it was updated. Merged! Thanks @oshalygin! 👍

oshalygin:
Thanks gents, love this starter btw :)

PR #393, opened 2017-03-16
Implement React Router@4.0.0 

TobiahRex:
Hi Cory, 

Here's your hot fresh PR. :) 

I've implemented react-router@4.0.0 under branch "react-router@4.0.0/tobiahrex".  

CHANGE NOTES:
- index.js:  
    * Implemented per the docs.  However I felt it cleaner to keep the Route tree out of the index.js file and in it's own component - App.js

- App.js:  Everything that was in "routes.js" is now converted (per my understanding of rrV4) and implemented here. 
   *  "IndexLink " is no longer an option. However "NavLink" with attr: "activeStyle" is.
   * "Switch" is for the NotFoundPage to catch anything not matched in the route tree, otherwise it's always mounted.
   * Each of the above mentioned comps are from react-router-dom since they're no longer a part of the rrV3 API. 

- configureStore.js: nothing outside of the docs except...
   * export const history = createHistory();  So the docs had the createStore() invoked in what is our "index.js" which I'm not a fan of.  Since we want the same "history" instance @ both the configureStore.js & index.js to pass in as "ConnectedRouter history={history}", I thought it best to do this.

- About, HomePage, NotFoundPage:
   * Added preceding "/" to each name. 
   * Changed import source to "react-router-dom".

- package.json: 
   * npm i history --save-dev 
   * npm i react-router 
   - react-router@4.0.0 is a package dependency so it's not explicitly installed.
   * npm i react-router-redux@next --save

Thanks for the suggestions to contribute and being open.  Let me know if I can do/change anything.

Respectfully, 

-Toby

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,10 +118,16 @@
     }
   },
   "babel": {
-    "presets": ["react", "stage-1"],
+    "presets": [
+      "react",
+      "stage-1"
+    ],
     "env": {
       "development": {
-        "presets": ["env", "react-hmre"]
+        "presets": [
+          "env",
+          "react-hmre"
+        ]
       },
       "production": {
         "presets": [
@@ -140,7 +146,9 @@
         ]
       },
       "test": {
-        "presets": ["env"]
+        "presets": [
+          "env"
+        ]
       }
     }
   },
@@ -151,7 +159,9 @@
       "plugin:import/errors",
       "plugin:import/warnings"
     ],
-    "plugins": ["react"],
+    "plugins": [
+      "react"
+    ],
     "parser": "babel-eslint",
     "parserOptions": {
       "ecmaVersion": 6,
@@ -173,7 +183,10 @@
       "no-console": 1,
       "no-debugger": 1,
       "no-var": 1,
-      "semi": [1, "always"],
+      "semi": [
+        1,
+        "always"
+      ],
       "no-trailing-spaces": 0,
       "eol-last": 0,
       "no-underscore-dangle": 0,
@@ -189,7 +202,9 @@
       "react/forbid-prop-types": [
         1,
         {
-          "forbid": ["any"]
+          "forbid": [
+            "any"
+          ]
         }
       ],
       "react/jsx-boolean-value": 0,
--- a/package.json
+++ b/package.json
@@ -112,10 +112,10 @@
     "url": "https://github.com/coryhouse/react-slingshot"
   },
   "jest": {
-    "moduleNameMapper": {
-      "\\.(css|scss)$": "identity-obj-proxy",
-      "^.+\\.(gif|ttf|eot|svg|woff|woff2|ico)$": "<rootDir>/tools/fileMock.js"
-    }
+    "moduleNameMapper": { 
+    "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/tools/assetsTransformer.js", 
+    "\\.(css|less)$": "<rootDir>/tools/assetsTransformer.js" 
+    }  
   },
   "babel": {
     "presets": [
--- a/package.json
+++ b/package.json
@@ -41,54 +41,54 @@
     "redux-thunk": "2.1.0"
   },
   "devDependencies": {
-    "autoprefixer": "6.5.4",
-    "babel-cli": "6.18.0",
-    "babel-core": "6.20.0",
-    "babel-eslint": "7.1.1",
-    "babel-jest": "18.0.0",
-    "babel-loader": "6.2.10",
-    "babel-plugin-transform-react-constant-elements": "6.9.1",
-    "babel-plugin-transform-react-remove-prop-types": "0.2.11",
-    "babel-polyfill": "6.20.0",
-    "babel-preset-env": "1.3.2",
-    "babel-preset-react": "6.16.0",
+    "autoprefixer": "7.1.2",
+    "babel-cli": "6.24.1",
+    "babel-core": "6.25.0",
+    "babel-eslint": "7.2.3",
+    "babel-jest": "20.0.3",
+    "babel-loader": "7.1.1",
+    "babel-plugin-transform-react-constant-elements": "6.23.0",
+    "babel-plugin-transform-react-remove-prop-types": "0.4.6",
+    "babel-polyfill": "6.23.0",
+    "babel-preset-env": "1.6.0",
+    "babel-preset-react": "6.24.1",
     "babel-preset-react-hmre": "1.1.1",
-    "babel-preset-stage-1": "6.16.0",
-    "browser-sync": "2.18.5",
-    "chalk": "1.1.3",
+    "babel-preset-stage-1": "6.24.1",
+    "browser-sync": "2.18.12",
+    "chalk": "2.0.1",
     "connect-history-api-fallback": "1.3.0",
-    "coveralls": "2.11.15",
-    "css-loader": "0.26.1",
-    "enzyme": "2.8.2",
-    "eslint": "3.12.2",
-    "eslint-plugin-import": "2.2.0",
-    "eslint-plugin-react": "6.8.0",
-    "eslint-watch": "2.1.14",
-    "extract-text-webpack-plugin": "2.1.0",
-    "file-loader": "0.9.0",
-    "html-webpack-plugin": "2.24.1",
+    "coveralls": "2.13.1",
+    "css-loader": "0.28.4",
+    "enzyme": "2.9.1",
+    "eslint": "4.2.0",
+    "eslint-plugin-import": "2.7.0",
+    "eslint-plugin-react": "7.1.0",
+    "eslint-watch": "3.1.2",
+    "extract-text-webpack-plugin": "3.0.0-rc.2",
+    "file-loader": "0.11.2",
+    "html-webpack-plugin": "2.29.0",
     "identity-obj-proxy": "3.0.0",
-    "jest": "18.1.0",
+    "jest": "20.0.4",
     "json-loader": "0.5.4",
     "mockdate": "2.0.1",
-    "node-sass": "4.5.2",
-    "npm-run-all": "3.1.2",
+    "node-sass": "4.5.3",
+    "npm-run-all": "4.0.2",
     "opn-cli": "3.1.0",
     "postcss-loader": "1.2.1",
     "prompt": "1.0.0",
     "prop-types": "15.5.10",
     "react-hot-loader": "3.0.0-beta.6",
     "react-test-renderer": "15.5.4",
-    "redux-immutable-state-invariant": "1.2.4",
+    "redux-immutable-state-invariant": "2.0.0",
     "replace": "0.3.0",
-    "rimraf": "2.5.4",
-    "sass-loader": "6.0.2",
-    "style-loader": "0.13.1",
-    "url-loader": "0.5.7",
-    "webpack": "2.2.1",
-    "webpack-bundle-analyzer": "2.1.1",
-    "webpack-dev-middleware": "1.9.0",
-    "webpack-hot-middleware": "2.17.1",
+    "rimraf": "2.6.1",
+    "sass-loader": "6.0.6",
+    "style-loader": "0.18.2",
+    "url-loader": "0.5.9",
+    "webpack": "3.1.0",
+    "webpack-bundle-analyzer": "2.8.2",
+    "webpack-dev-middleware": "1.11.0",
+    "webpack-hot-middleware": "2.18.2",
     "webpack-md5-hash": "0.0.5"
   },
   "keywords": [
@@ -117,16 +117,10 @@
     }
   },
   "babel": {
-    "presets": [
-      "react",
-      "stage-1"
-    ],
+    "presets": ["react", "stage-1"],
     "env": {
       "development": {
-        "presets": [
-          "env",
-          "react-hmre"
-        ]
+        "presets": ["env", "react-hmre"]
       },
       "production": {
         "presets": [
@@ -145,9 +139,7 @@
         ]
       },
       "test": {
-        "presets": [
-          "env"
-        ]
+        "presets": ["env"]
       }
     }
   },
@@ -158,9 +150,7 @@
       "plugin:import/errors",
       "plugin:import/warnings"
     ],
-    "plugins": [
-      "react"
-    ],
+    "plugins": ["react"],
     "parser": "babel-eslint",
     "parserOptions": {
       "ecmaVersion": 6,
@@ -182,10 +172,7 @@
       "no-console": 1,
       "no-debugger": 1,
       "no-var": 1,
-      "semi": [
-        1,
-        "always"
-      ],
+      "semi": [1, "always"],
       "no-trailing-spaces": 0,
       "eol-last": 0,
       "no-underscore-dangle": 0,
@@ -201,9 +188,7 @@
       "react/forbid-prop-types": [
         1,
         {
-          "forbid": [
-            "any"
-          ]
+          "forbid": ["any"]
         }
       ],
       "react/jsx-boolean-value": 0,
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "html-webpack-plugin": "2.29.0",
     "identity-obj-proxy": "3.0.0",
     "jest": "20.0.4",
+    "jest-cli": "20.0.4",
     "json-loader": "0.5.4",
     "mockdate": "2.0.1",
     "node-sass": "4.5.3",
--- a/package.json
+++ b/package.json
@@ -136,8 +136,12 @@
             {
               "es2015": {
                 "modules": false
+              },
+              "targets": {
+                "ie": 9,
+                "uglify": true
               }
-            }
+           }
           ]
         ],
         "plugins": [
--- a/package.json
+++ b/package.json
@@ -35,9 +35,9 @@
     "react": "15.5.4",
     "react-dom": "15.5.4",
     "react-redux": "5.0.5",
-    "react-router": "3.0.5",
+    "react-router": "4.2.0",
     "react-router-dom": "4.0.0",
-    "react-router-redux": "4.0.8",
+    "react-router-redux": "5.0.0-alpha.6",
     "redux": "3.6.0",
     "redux-thunk": "2.1.0"
   },
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "react-dom": "15.5.4",
     "react-redux": "5.0.5",
     "react-router": "3.0.5",
+    "react-router-dom": "4.0.0",
     "react-router-redux": "4.0.8",
     "redux": "3.6.0",
     "redux-thunk": "2.1.0"
@@ -66,6 +67,7 @@
     "eslint-watch": "3.1.2",
     "extract-text-webpack-plugin": "3.0.0-rc.2",
     "file-loader": "0.11.2",
+    "history": "4.6.0",
     "html-webpack-plugin": "2.29.0",
     "identity-obj-proxy": "3.0.0",
     "jest": "20.0.4",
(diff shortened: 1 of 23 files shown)


TobiahRex:
Anyone have any clues on getting those checks to pass?

nickytonline:
@TobiagRex, as the logs state there were some network issues. I know that 
Yarn was down yesterday for some reason. Probably relaunch the  builds to see if you still get the issue. Also, please add the new react router packages to the yarn.lock file. See  https://yarnpkg.com/en/docs/yarn-lock. You can add it via `yarn add -E`

mikedevita:
Thanks for adding the spacing around the `{` and `}` that always bugged me lol.

TobiahRex:
@nickytonline Good Feedback! Saw the logs but didn't want to assume eithout peer input since I have little exposure to yarn.  The node version change log entry also made me double check my own node version and alas I was out of date (since updated).

Also copy all @ yarn.lock file.  Will add react router package changes and re-commit ASAP.

break...

@mikedevita haha yeah I'm a stickler for that stuff.  All thanks to eslint w/airbnb style guide extension.

TobiahRex:
@all, apologies for the typos...

"reach" = each*


"npm i react-router --save" = 
npm i react-router-dom* --save
- react-router@4.0.0 is a package dependency so it's not explicitly installed

src/store/configureStore.js
```diff
@@ -30,6 +36,9 @@ function configureStoreDev(initialState) {
     thunk,
   ];
 
+  const reactRouterMiddleware = routerMiddleware(history);
```
oshalygin: I wouldn't push to the array, instead declare the array below all this and just include it ineline.

package.json
```diff
@@ -35,8 +35,8 @@
     "react": "15.4.1",
     "react-dom": "15.4.1",
     "react-redux": "5.0.1",
-    "react-router": "3.0.0",
-    "react-router-redux": "4.0.7",
+    "react-router-dom": "4.0.0",
+    "react-router-redux": "5.0.0-alpha.3",
```
oshalygin: This feels a bit rough to be using an alpha package.  Many of us are using this library for projects and these things will inevitably break into beta+

coryhouse:
Thanks so much for the PR! 👍 💯  I agree that we need to wait until all deps are out of alpha. But this PR is an excellent foundation once react-router-redux goes 5 final releases.

TobiahRex:
@nickytonline & @coryhouse  Understood.  Many thanks for the feedback.

kwelch:
*What is the need/use-case for react-router-redux? Why include it?*

It understand it adds the location to the store, but what benefit does that have? Does it still warrant being included in slingshot?

The only valid case I have for this at this point is redux dev tools time travel. 

I ask not to try and push this through, but more so to ensure we asked and answered the question. 

This was something I chose to remove when I move our company projects to RRv4.

TobiahRex:
@kwelch Great question.  I see 2 benefits: 1) We can dispatch session history changes via redux actions - I would argue that it's "cleaner".  2) The most powerful use case I could think of is it would also allow us to cancel asynchronous view changes when implemented with Observables that can take advantage of redux-middleware.

https://github.com/redux-observable/redux-observable

TobiahRex:
@kwelch Great question.  Outside of a dev environment, i'm not sure there's a benefit of react-router-redux.  

The guys' description @ it's npm page even says...
> This library is not necessary for using Redux together with React Router. You can use the two together just fine without any additional libraries. It is useful if you care about recording, persisting, and replaying user actions, using time travel. If you don't care about these features, just use Redux and React Router directly.

https://github.com/redux-observable/redux-observable

kwelch:
@coryhouse / @nickytonline  How do you feel about dropping react-router-redux?

nickytonline:
@kwelch, I've never used it so I'll let you and @coryhouse make that call.

TobiahRex:
@nickytonline update on upgrading the yarn.lock file...

I'm currently in Japan on a freelance job and will be for few more months.  I've attempted to update the yarn.lock file with the upgrades to react-router version numbers & add react-router-dom & react-router-redux.  Problem is, all 3 locations I've attempted to do this from, are giving me 

`"There appears to be trouble with your network connection. Retrying..."` 

errors from the yarn cli. After some research, (https://github.com/yarnpkg/yarn/issues/668) it appears to be a common issue if the internet bandwidth is less than 2-2.5mb/s.  Alas, this is my case here in Japan from all 3 locations.  So, I'll continue to try, however someone stateside may have a faster time upgrading the yarn.lock file.  I'll re-commit as soon as I am successful.

src/store/configureStore.js
TobiahRex: @oshalygin 
Fair enough. I'll change it.

kwelch:
Thanks @TobiahRex, as long as you have opened write access to maintainers, we should be able to push a commit to fix the lock file.

TobiahRex:
FYI: I decided to try complete removal of node_modules/ and then run `yarn` install to sneak around the "Network Connection" error I mentioned previous, as well as having the added benefit of yarn linking dependencies and making my post yarn install node_modules/ dir much smaller. 

Result: 
1) yarn.lock file updated react-router + friends! 
2) yarn.lock file also updated all packages listed.

nickytonline:
@TobiahRex, looks like there's conflicts. Also, CI is failing with your latests commit.

TobiahRex:
@nickytonline Forgot to rebase master & PR branch from the latest PR merges.  In progress now.  Will re-commit ASAP.

package.json
```diff
@@ -78,7 +78,7 @@
     "postcss-loader": "1.2.1",
     "prompt": "1.0.0",
     "react-addons-test-utils": "15.4.1",
-    "react-hot-loader": "^1.3.1",
+    "react-hot-loader": "3.0.0-beta.6",
```
oshalygin: Actually latest has 3.0.0-beta.6.  I dun goofed and had `3.0.0` which isn't an actual version that can be pulled down.

If you guys would like I can PR us back(revert HMR 3.0).  LMK @coryhouse  and my apologies.

coryhouse: Yeah, now that we've already pulled in 3.0.0-beta.6, let's just leave it. It's been out there for a few days without anyone reporting issues and I've been told it's quite stable. Not ideal, but at this point I'd say it's better to leave it than churn further on a rollback. Thanks for your work on the RR4 conversion!

package.json
coryhouse: Agreed. I'd like to avoid taking on alpha/beta dependencies.

package.json
nickytonline: Looks like we're good to go then.

package.json
nickytonline: Ahh didn't see this one here still. @TobiahRex, can the RR4 PR use a stable version of `react-router-redux`?

coryhouse:
Torn on how to handle this PR. I agre that we shouldn't take a dependency on an alpha version of react-router-redux. Options:
1. Drop react-router-redux for now
2. Wait for the final release.

I lean toward 2.

Suggestions @nickytonline @kwelch?

nickytonline:
@coryhouse, waiting for the final release works for me.

package.json
TobiahRex: @nickytonline To answer the question, Yeah RR4 works with react-hot-loader 1.3.1.
@oshalygin That makes sense.  I was wondering...
@coryhouse Happy to contribute!! :)

TobiahRex:
Apologies for the commits - wanted to try RR4 with react-router-redux v4 and in doing so found a small bug in my previous "successful commit".  It's been squashed. & RRRx4 + RR4 = No Go.

src/components/Root.js
```diff
@@ -1,20 +1,22 @@
...
+        </ConnectedRouter>
       </Provider>
     );
   }
 }
 
 Root.propTypes = {
   store: PropTypes.object.isRequired,
-  history: PropTypes.object.isRequired
+  history: PropTypes.object.isRequired,
```
oshalygin: consider removing trailing comma

src/components/App.js
```diff
@@ -1,20 +1,37 @@
 import React, { PropTypes } from 'react';
-import { Link, IndexLink } from 'react-router';
+import { Route } from 'react-router';
+import { Switch, NavLink } from 'react-router-dom';
+// 'Switch' & 'NavLink' imports have moved from the 'react-router' API to 'react-router-dom' since rrV4;
+// Also, IndexLink is replaced by NavLink with the activeStyle attr.
+// Switch is required for NotFoundPage to be the last case scenario.
+
+import HomePage from './HomePage';
+import FuelSavingsPage from '../containers/FuelSavingsPage'; // eslint-disable-line import/no-named-as-default
```
oshalygin: consider removing this eslint-disable, your code meets the import/no-named-as-default rule afaict.

src/components/Root.js
TobiahRex: @oshalygin lol, i seem to be the only one who hasn't disabled that rule in .eslintrc files.  Newest commit has both changes you suggested. Thanks for the feedback!

kwelch:
I am fine waiting. I think that is simpler than explaining a drop in support for a library.

gudh:
How about replace react-router-redux (which is still in alpha) with [connected-react-router](https://github.com/supasate/connected-react-router)

package.json
gudh: How about replace react-router-redux (which is still in alpha) with [connected-react-router](https://github.com/supasate/connected-react-router) ?

ozasadnyy:
How to make `<ConnectedRouter>` working with `basename` as `<BrowserRouter>` [do](https://reacttraining.com/react-router/web/api/BrowserRouter/basename-string)?

kwelch:
I have not used ConnectedRouter, so I am unsure and from a brief look at the code it does not appear to accept that param. It may be a good question for StackOverflow or a bug report for react-router

edaloicaro18:
I want to use the slingshot, but from what I see it is still using react-router v3.

oshalygin:
@coryhouse @TobiahRex let me know if I can help guys, I'm just about to roll RR4 into my side project

TobiahRex:
@oshalygin, @edaloicaro18 We agreed to wait for react-router-redux v5 to come out of alpha before merging this pr. The status as of today is still 'alpha'.

https://github.com/ReactTraining/react-router/blob/master/packages/react-router-redux/package.json

TobiahRex:
@kwelch & @nickytonline Rebase complete. Couple notes...
1) Coveralls Coverage percentage dropped a tad (.7%).  I looked at the report and it doesn't appear to be coming from any files I personally changed (see image).  I could be wrong, but it looks like it's coming from the package dependencies version upgrades.  Would need another to confirm.
screen shot 2017-08-31 at 3 18 59 pm


2) I believe `react-router-redux` hasn't upgraded their Prop Validation library to `prop-types` post `react v15` (their v5 is still in "alpha") hence this error ... PropType validation error from npm Package  I've triple checked all our components for proper `prop-types` lib implementation and validation and we're 100% GTG.

package.json
```diff
@@ -35,60 +35,63 @@
     "react": "15.5.4",
     "react-dom": "15.5.4",
     "react-redux": "5.0.5",
-    "react-router": "3.0.5",
-    "react-router-redux": "4.0.8",
+    "react-router": "4.2.0",
+    "react-router-dom": "4.0.0",
+    "react-router-redux": "5.0.0-alpha.6",
```
kwelch: Just want to ask this again before we move forward...

Are we sure we want to keep this?

kwelch:
The router changes look good. I am nervous that it shows so much diff in other areas. I would not have expected that after a rebase.

package.json
nickytonline: cc: @coryhouse

src/components/AboutPage.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import {Link} from 'react-router';
+import { Link } from 'react-router-dom';
```
nickytonline: which formatting does our linter support? Spacing between {} in imports or none? I can't remember. cc: @kwelch.

src/components/AboutPage.spec.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import {shallow} from 'enzyme';
+import { shallow } from 'enzyme';
```
nickytonline: same as above.

src/components/FuelSavingsForm.js
```diff
@@ -32,48 +32,48 @@ class FuelSavingsForm extends React.Component {
...
-                <option value="month">Month</option>
-                <option value="year">Year</option>
-              </select>
-            </td>
-          </tr>
-          <tr>
-            <td><label>Date Modified</label></td>
-            <td>{fuelSavings.dateModified}</td>
-          </tr>
+            <tr>
```
nickytonline: Once again, question about the spacing changes. Is this our eslint config doing this?

nickytonline:
@kwelch, it looks like a lot of the changes are spacing changes.

src/components/AboutPage.js
kwelch: Not sure, I know I prefer with spaces. I also believe prettier default is with spaces.

kwelch:
Good point, that may be prettier related. I don't think we currently are `prettier` friendly, but I would not mind us moving that direction. 👍

kwelch:
I think this is all set with one outstanding questions see below. 

OPEN QUESTION: Do we want to continue to use/support `react-router-redux`? 
It is still in alpha.

TobiahRex:
@kwelch & @coryhouse 
You all may already know this, but for the sake of clarity and making an informed decision on whether to continue waiting or dropping the alpha version, i'd like to shed some light on what purpose `react-router-redux 5-alpha` is serving.

1) The `ConnectedRouter` component import is exclusive to `react-router-redux 5-ALPHA` (unavailable in `react-router-redux 4`).  It's purpose (if you don't already know) is to keep a record of route changes in redux. https://github.com/ReactTraining/react-router/blob/master/packages/react-router-redux/modules/ConnectedRouter.js#L19

2) Part of `rr4`'s new version was to cake in this redux storage & implementation of route changes automatically.  The dispatching happens via the `push` func via redux - previously it was done `browserHistory.push(<path>)` if you recall.  https://github.com/ReactTraining/react-router/blob/master/packages/react-router-redux/modules/actions.js#L21

So my point is, if we remove the `react-router-redux 5-ALPHA` we're removing dispatchable actions such as `push` and redux storage automatically.  It's not a requirement for rr4 usage as long as the dev implements some alternative route change api - global browser vars like `window.location`.

On a personal note, I find using redux's route history storage highly convenient.  A couple use cases of mine:
1) If user has navigated to a nested route within the app, then logs into their account, this will typically force a browser reload, defaulting to the homepage post login.  With something like `redux-persist` you can cache the redux stores route branch pre-login, then forward the user to that cached route after successful login.
2)  Part of `react-rotuer-redux` default behavior is to also save things query params automatically.  This is pretty useful if you have re-direct links (from an email as an example) to a client-side page, that needs to pass a token back to the server to fetch some data.
Redux branch that's added from `react-router-redux`


Overall i think the benefit is simply not having to access a global browser api (window) for extracting /mutating route data.

src/components/FuelSavingsForm.js
TobiahRex: @nickytonline This one looks like its from my text-editor.  I use an atom package `language-babel` to automatically calculate JSX spacing.  In this particular case, i think nudged it over accidentally.

oshalygin:
Just for reference, I tried to roll in `react-router-redux` and I was seeing strange bugs left and right with routing(mostly related to HMR).  It could be my own codebase and this is totally anecdotal but I would love to see full support before rolling it in.

I'm open to using something that works and if people have an alternative I'd love to hear it and I'd love to try it out and let you know if it works.

Sidenote:
After upgrading to RR4 I'm not sure what I gained.  I feel like its just a different mental model to manage within the application.  I get that things are moving to the new way of things and leveraging Reacts lifecycle hooks instead of RR's, but I digress...

kwelch:
@TobiahRex 
Thank you for the added explanation. 

I, like @oshalygin, found myself in a different mental model for routes after RRv4 and did not require the ability to push from within an action. I understand that I am not the only voice, I also don't want to introduce and support something that is in alpha as the purpose of slingshot is to make entry into the React ecosystem more approachable. 

I am good with either approach. My main objective is to ensure this project remains stable, friendly, and approachable to all.

package.json
coryhouse: I vote for dropping it. Easy to add if desired, and avoids an alpha dep

src/components/Root.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { ConnectedRouter } from 'react-router-redux';
+console.log('%cConnectedRouter', 'background:red;', ConnectedRouter);
```
coryhouse: Looks like a console.log slipped in

src/components/FuelSavingsForm.js
coryhouse: @TobiahRex - Would you be willing to rollback the whitespace changes in this PR to keep the diff minimal?

src/store/configureStore.js
```diff
@@ -1,15 +1,20 @@
 import {createStore, compose, applyMiddleware} from 'redux';
 import reduxImmutableStateInvariant from 'redux-immutable-state-invariant';
 import thunk from 'redux-thunk';
+import createHistory from 'history/createBrowserHistory';
+// 'routerMiddleware': the new way of storing route changes with redux middlware since rrV4.
```
coryhouse: Missing "e" in middleware

coryhouse:
@TobiahRex - Thanks for the thoughtful input on react-router-redux. I'm torn on whether to include it given its alpha status and the ease with which it can be added in - but I'm good with leaving it since it's already in here. I made some simple requests for changes. Thanks so much for the hard work on the PR Tobiah!

TobiahRex:
@coryhouse - Saw and fixed issues per your feedback.  Thanks for the review.  Understood on being torn.  Happy to contribute and I'll continue re-basing until we're ready to merge.

coryhouse:
@TobiahRex - Looks like FuelSavingsForm.js still has unnecessary whitespace changes. Can you please revert?

TobiahRex:
@coryhouse I left a note asking for confirmation before reverting. I think the white space you're referring to is correct due to hierarchical precedence of `<tbody>` to a `<tr>`.  Please confirm.

src/components/FuelSavingsForm.js
```diff
@@ -32,48 +32,48 @@ class FuelSavingsForm extends React.Component {
         <h2>Fuel Savings Analysis</h2>
         <table>
           <tbody>
-          <tr>
```
TobiahRex: @nickytonline & @coryhouse 
I believe the white spacing change I made here is correct. Should not a `<tr>` appear as a child of a `<tbody>` in this case and not a sibling?  Please confirm you'd like it to appear as a sibling and I'll revert.

src/components/FuelSavingsForm.js
TobiahRex: @nickytonline & @coryhouse Please see note above ☝️

TobiahRex: @nickytonline & @coryhouse After checking again, i think this is a correct addition of whitespace.  The `<tr>` should appear as a **child** of `<tbody>` and not a **sibling** due to hierarchical precedence, correct?  If you'd sitll like me to revert back to what it was before, i can do that. please confirm.

src/components/FuelSavingsForm.js
coryhouse: Point taken. You can leave it.

coryhouse:
I'm good w/ merging this. Any remaining reservations? @kwelch @nickytonline

kwelch:
Not see anything out of the ordinary. LGTM

nickytonline:
LGTM too.

src/components/FuelSavingsForm.js
TobiahRex: @coryhouse thanks for the confirmation.

coryhouse:
Merged! 🎉 Thanks so much for your hard work @TobiahRex!

TobiahRex:
@coryhouse Thank you! Back when I first started learning React, the slingshot repo and your plural sight course was my first exposure.  Really happy to contribute.

PR #401, opened 2017-03-30
Updated react-hot-loader to correct package version.

danielcampo:
`npm install` would fail to execute otherwise.


coryhouse:
Sigh. I don't know why the PR was updated to reference a non-existent version. Guess we'll live with the beta. Thanks for pointing this out. 👍

danielcampo:
Chalk it up to another one of life's mysteries... And my pleasure!

PR #405, opened 2017-04-04
Move babel and eslint configuration to package.json

coryhouse:
It's easy to overlook .babelrc and .eslintrc when copying files, especially on Windows. So I think it makes sense to move them to package.json. Many other boilerplates use this approach as well.

What does everyone think?


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-slingshot",
-  "version": "5.0.0",
+  "version": "7.0.0",
```
kwelch: Why the big version hike?

nickytonline:
I'm fine with the change as well. Might as well stop causing headaches for Windows users.

package.json
coryhouse: Well embarrassingly, when I announced 6.0, I didn't update package.json to reflect it. I would consider the config move a breaking change, so that's why I went to 7. Make sense?

coryhouse:
The size of package.json wouldn't be bad if we dropped all the custom eslint rules. I'd like to consider dumping all custom rules and just use eslint's recommended as a basline. Thoughts on that?

nickytonline:
I'm OK with that.


On Tue, Apr 4, 2017 at 9:36 AM Cory House <notifications@github.com> wrote:

*@coryhouse* commented on this pull request.
------------------------------

In package.json
<https://github.com/coryhouse/react-slingshot/pull/405#discussion_r109663120>
:

> @@ -1,6 +1,6 @@
 {
   "name": "react-slingshot",
-  "version": "5.0.0",
+  "version": "7.0.0",

Well embarrassingly, when I announced 6.0, I didn't update package.json to
reflect it. I would consider the config move a breaking change, so that's
why I went to 7. Make sense?

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/coryhouse/react-slingshot/pull/405#discussion_r109663120>,
or mute the thread

kwelch:
I agree for the purposes of this repo a baseline/standard rule list may be better suited. 

Any updates we can make to the babel config to make that cleaner? Maybe introduce `babel-preset-env`.

package.json
kwelch: Yeah, I forgot the current release was 6.0, so I was mostly trying to figure how one change was 2 version. All good by me.

nickytonline:
@coryhouse, I'll take a peek tonight at your PR.

package.json
nickytonline: same here.

nickytonline:
@kwelch anything else to add before I merge?

kwelch:
I am good, I think there are few action items from this around the lint and babel configs. Should we open an issue for those to discuss?

nickytonline:
@kwelch, yeah, can you open another issue for that?

PR #409, opened 2017-04-06
Fix: Regex to replace keywords on running setup

dineshvgp:


coryhouse:
Thanks! Can you clarify the issue that was occurring with the previous regex?

dineshvgp:
@coryhouse It finds till last occurrences of `]`. Instead, it should replace only keywords array.

This isssue occurs only after moving eslint and babel to package.json. Only in those properties we are using `]`

coryhouse:
Ah, good catch! Thanks for clarifying and for the PR! 👍

PR #411, opened 2017-04-07
Fix: Module deprecated warnings

dineshvgp:


coryhouse:
Thanks! 👍

kwelch:
Looks like I didn't have to make the issue about switching to env. Change look great!

Issue #413, opened 2017-04-08
Update the issue template with things to try first.


nickytonline:
@coryhouse, @kwelch, this is a small PR to add a things to try before submitting an issue in the issue template. Let me know if the wording is OK or if you'd like some tweaks to it.

Issue #326, opened 2016-11-22
Issue with serving static files


jediofthecode:
Node version: 6.9.1
npm version: 3.10.8
Operating system: Ubuntu 16.10 AMD64

Steps to reproduce: When serving fonts via the file loaders in src/fonts, they are available to download, and if set with an absolute http url (eg: http://localhost:3030/fonts/fontname.woff in sass they are correctly read, but if I set with an absolute base url(eg: /fonts/fontname.woff) it will not load properly. I am wondering if this is an issue or there are special steps for including fonts in sass files for defining custom font-faces.

kwelch:
Is this from the dev server or production? 

If you are intending to use the file loader I recommend referencing the files like the [image example on the FAQ page](https://github.com/coryhouse/react-slingshot/blob/master/docs/FAQ.md#how-do-i-handle-images).

jediofthecode:
it is from the dev server, and, can I reference files like this within an scss file ?

davidglbr:
Yes, if I understand what you are asking.
Can you describe the use case so maybe we can be more helpful?
ᐧ

jediofthecode:
Alright, sorry for the delay, if I reference a font like this in my sass:

`@font-face {
    font-family: 'proxima-reg';
    src: url('../fonts/proxima_nova_alt_regular-webfont.woff2') format('woff2'),
         url('../fonts/proxima_nova_alt_regular-webfont.woff') format('woff');
    font-weight: normal;
    font-style: normal;
}`

it will not work at first, it shows no font's loading, i get no errors in webpack or in the console. Then i can modify the file to anything, save, and modify it back to exactly this, save and it hotloads, and the font loads...this makes no sense why it is not loading initially.

davidglbr:
Weird. That worked for me. Maybe it's something with your specific
hot-reloading config? Have you changed it from the one on the repo?
ᐧ

kwelch:
I put in #207 which dynamically set the `publichPath` which should resolve this. 

If I read your last comment it appears you got it working. If not feel free to re-open.

PR #414, opened 2017-04-08
Added things to try before reporting an issue to the template.

nickytonline:
This implements #413.


nickytonline:
The steps to try before logging an issue seem obvious to me, but we've had several issues reported where it appears a user did not do these things first, so I think it's better to be explicit on what to try first instead of us asking these things in the issues they report, e.g. https://github.com/coryhouse/react-slingshot/issues/326#issuecomment-262254378

coryhouse:
Love it. Looks great! Thanks so much Nick!

PR #418, opened 2017-04-10
migrate from React.PropTypes

gabrielkunkel:
This may be a bit early, but it has to be done some time...

With React 15.5.0, [React.PropTypes are now extracted to a separate package](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes).

They are still accessible via the main React object, but using them will log a deprecation warning to the console when in development mode.
3390e426-1c7c-11e7-8ee3-28720e13eb99


coryhouse:
Thanks! 👍

mikedevita:
I'm curious, why is there another lib for `prop-types`? Is that now not included in react?

coryhouse:
It was a change in 15.5

Issue #395, opened 2017-03-19
open:cover - no open on windows


deanbot:
Greetings.

Node version: 6.10.0

npm version: 3.10.10

Operating system: Windows

Command line used: npm run open:cover

Steps to reproduce: when it gets to the `open coverage/lcov-report/index.html` the error is "'open' is not recognized as an internal or external command, operable program or batch file."

nickytonline:
@deanbot, I'll ask the obvious first. Did you execute `npm install` or `yarn` before running `npm run open:cover`? The error you have seems to indicate the [npm package for `open`](https://github.com/coryhouse/react-slingshot/blob/master/package.json#L76) is not installed.

deanbot:
@nickytonline 

Yup:

```
rm -rf node_modules
yarn
yarn run open:cover
```

I'm not sure that the open npm project provides a cli command though I could be wrong. Note examples on https://github.com/pwnall/node-open don't show cli commands. See this earlier iteration using open https://github.com/coryhouse/pluralsight-redux-starter/blob/master/tools/srcServer.js#L28

This works:

package.json:
`"open:cover": "npm run test:cover && babel-node tools/openCover.js",`

and in tools/openCover.js add:
```
import open from 'open';
import path from 'path';
open(path.resolve(__dirname, '../coverage/lcov-report/index.html'), 'chrome');
```

Unfortunately specifying 'chrome' is necessary since my default app for .html is an IDE.

I think the better fix would be to serve it over browser-sync.

nickytonline:
@deanbot, sorry for the delayed response. I started a new job recently. I'm not sure we want to spin up browser-sync because the default app for HTML files is not a browser. @kwelch, @coryhouse... thoughts?

coryhouse:
I see your point about open not appearing to support a command line call. Not sure when this was regressed, but would appreciate a PR to call a separate script.

However, we can't specify a specific browser like you did since there's no guarantee a given browser would exist.

PR #421, opened 2017-04-17
add openCover script to support open:cover command on windows

deanbot:
see issue #395


RoM4iK:
Then, we can move to a better opener package, instead of writing wrappers for existing.
We can look at [opn](https://github.com/sindresorhus/opn), and exactly on [opn-cli](https://github.com/sindresorhus/opn-cli)

coryhouse:
Ah good idea @RoM4iK. I wasn't aware of opn-cli, but it looks perfect. @deanbot - Would you be willing to adjust your PR to use it?

PR #424, opened 2017-04-25
Actually copy the favicon to dist/ on prod build

mikedevita:
this helps with #128 (even though it was closed).


kwelch:
Is the update to the `.ejs` required? my only concern is the public path not being set to what you have hard coded.

mikedevita:
@kwelch i don't think its explicitly required, I just added it so others have one less step to add. Is there a more ideal way to include it? maybe more along the lines of `<link rel="icon" href="favicon.ico">`

Issue #408, opened 2017-04-06
Add Deployment Section to Docs


kwelch:
React Slingshot has a solid documentation on how to start developing, testing, and building a React/Redux application, but is missing a fundamental piece, deployment. 

As we all know it doesn't matter what you build if you never ship. This issue will give visibility to this need and also allow us to gather and prioritize the deployment options covered. 

Please reply with deployment plans you would like to see so we can vote on them. Additionally, if you have experience with a deployment plan and would like to document it please inform us so no work is duplicated.

rizkysyazuli:
I've deployed using an Express server as discussed in #263. works pretty well although still in a staging environment. +1 for Express.

aloisbarreras:
I've deployed using an S3 static website. I'm happy to write documentation for it.

coryhouse:
Hi @aloisbarreras - I'd welcome a PR on that in the README.md. Thanks!

PR #428, opened 2017-05-04
Add express to serve static files

### LINKED ISSUE #427: Add express to serve static files

iLeonelPerea:
closes #427 

Changes proposed in the pull request:

- Add express for production
- Update favicon

@coryhouse


kwelch:
Could you provide the 'why' for this change? What does this give us?

iLeonelPerea:
@kwelch

The actual configuration uses browsersync, isn't intended for production, to serve static files and i propose to use express. Because the actual configuration of browsersync share the same sesion if you access from multiple web browsers. 

Maybe some developers try to use the same configuration for production that does not work with browsersync.. 

But I got your point..

ryannealmes:
Surely where the files from production get served is separate from this repo? For example I build react-slingshot and store my files in an S3 bucket and serve it from there. Express wouldn't really serve any purpose.

kwelch:
Yes, the intent of the distServer is to validate the production code. Its purpose was not to be a production ready server. This may server better as part of the documentation we are looking to compile for #408

PR #443, opened 2017-07-10
Add redux-logger to dev store configuration

arsood:
This change adds the redux-logger package to devDependencies and also to the Redux dev store configuration for better debugging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "react-hot-loader": "3.0.0-beta.6",
     "react-test-renderer": "15.5.4",
     "redux-immutable-state-invariant": "1.2.4",
+    "redux-logger": "3.0.6",
     "replace": "0.3.0",
     "rimraf": "2.5.4",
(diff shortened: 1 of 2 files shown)


kwelch:
I don't see an issue with this directly. Not sure if we need redux-logger since the redux dev tools (which is already integrated) covers this need. Additionally, if we were to accept this there are some issues with indentation.

coryhouse:
Hi @arsood - Thanks for the PR.

However, I would prefer to leave this out since:
1. This is easy to add
2. Redux dev tools covers this area too
3. It's trivial to add `debugger` where desired as well.

But thanks for the thoughtful PR! :)

arsood:
@coryhouse thanks for the note! I found it helpful while using your boilerplate so I thought I'd send in a PR in case you found it useful.

Thanks for making the boilerplate! It has been very valuable for my React work :)

PR #444, opened 2017-07-10
Upgrade to webpack 3, and upgrade all other dependencies

### LINKED ISSUE #442: Upgrade to Webpack 3
Webpack 3 was released this week, with new features like scope hoisting, https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b .We should let the dust settle a bit, but in a week or two, we should consider upgrading.

This is probably a good first PR as 
 well since, as they say in the post, **"Migrating from webpack 2 to 3, should involve no effort beyond running the upgrade commands in your terminal."**

egdelwonk:
Haven't done a ton of testing on this branch yet, but so far it looks ok. 

This will upgrade to **Webpack 3** and _upgrades a ton of other packages_ too. If you'd like, I can try to separate those dependency upgrades out, but figured it'd be useful to get everything up to the latest versions.

Resolves #442

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,54 +41,54 @@
     "redux-thunk": "2.1.0"
   },
   "devDependencies": {
-    "autoprefixer": "6.5.4",
-    "babel-cli": "6.18.0",
-    "babel-core": "6.20.0",
-    "babel-eslint": "7.1.1",
-    "babel-jest": "18.0.0",
-    "babel-loader": "6.2.10",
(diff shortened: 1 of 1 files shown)


kwelch:
Odd that coverage would go down for an non-code change. I will pull this down and test it tonight. 

Thanks for the PR, Will!

nickytonline:
Related, #442.

coryhouse:
I made a couple minor tweaks, but looked great. Thanks @egdelwonk 👍

PR #450, opened 2017-07-18
Issue #449 fix

DaneTheory:
Ref: #449

Issue 1:
Removed extra dashes located in the package.json test scripts that cause start script to fail.

Issue 2:
Removed escape characters found in setupPrompts.js which cause linting to fail, thus breaking start script.

PR #451, opened 2017-07-18
Revert PR #450

DaneTheory:
Removed change that removed additional dashes in npm test scripts in favor of adding jest-cli as a devDep.

This commit instead focusses on issue 2 from #449 where setupPrompts.js had a bug that caused start script to fail.

### DIFF
--- a/tools/setup/setupPrompts.js
+++ b/tools/setup/setupPrompts.js
@@ -3,7 +3,7 @@ module.exports = [
   {
     name: 'projectName',
     description: 'Project name (default: new-project)',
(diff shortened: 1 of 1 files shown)


DaneTheory:
@coryhouse here's the revert for #450 

Couldn't get the revert to work via CLI so just did a manual fix. At work, otherwise I'd do it better.

coryhouse:
Thanks! 👍

kwelch:
PR #450 never merged so why did we need to revert?

PR #452, opened 2017-07-21
Enhance babel env config to transpile for IE9+

coryhouse:


kwelch:
I don't see any issue with this. I haven't bug very deep into `env` so all appears right from the surface.

PR #466, opened 2017-09-03
Change build and clean to get target from package.json

raythree:
Changed build and clean to get target configuration and webpack public paths from package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,8 +18,9 @@
     "lint:watch": "npm run lint -- --watch",
     "clean-dist": "npm run remove-dist && mkdir dist",
     "remove-dist": "rimraf ./dist",
-    "prebuild": "npm run clean-dist && npm run lint && npm run test",
-    "build": "babel-node tools/build.js && npm run open:dist",
+    "prebuild": "npm run clean && npm run lint && npm run test",
+    "build": "babel-node tools/build.js",
+    "clean": "babel-node tools/clean.js",
     "test": "jest",
     "test:CI": "babel-node tools/testCi.js",
     "test:cover": "npm run test -- --coverage ",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -18,8 +18,9 @@
     "lint:watch": "npm run lint -- --watch",
     "clean-dist": "npm run remove-dist && mkdir dist",
```
m10: clean-dist should be removed

tools/clean.js
```diff
@@ -0,0 +1,32 @@
...
+const configuredTarget = path.resolve(config.output.path);
+
+// Don't clean files that are "too close" to root. By default it is expected
+// that the project output would be at least 3 levels below root, like
+// /projects/app/dist. This value can be changed by setting the minCleanDepth
+// in the config section of package.json.
+const minCleanDepth = process.env.npm_package_config_minCleanDepth || 3;
+
+let depth = 0;
+configuredTarget.split('/').forEach(function (part) {
```
m10: this will break with windows

m10: .split(/\/|\\/) should do

webpack.config.prod.js
```diff
@@ -6,6 +6,11 @@ import WebpackMd5Hash from 'webpack-md5-hash';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import path from 'path';
 
+const configuredTarget = process.env.npm_package_config_targetPath ? 
+    path.resolve(process.env.npm_package_config_targetPath) :
+    path.resolve(__dirname, 'dist');
+const publicPath = process.env.npm_package_config_publicPath || '/';
```
m10: does not allow publicPath to be configured as empty string (for relative urls)

tools/build.js
```diff
@@ -29,7 +29,8 @@ webpack(config).run((error, stats) => {
   console.log(`Webpack stats: ${stats}`);
 
   // if we got this far, the build succeeded.
-  console.log(chalkSuccess('Your app is compiled in production mode in /dist. It\'s ready to roll!'));
+  console.log(chalkSuccess('Your app is compiled in production mode in ' + 
+    config.output.path + '. It\'s ready to roll!'));
```
m10: prefer template strings

m10:
it would be nice if these configs would be asked during setup

Issue #467, opened 2017-09-04
Webpack config target: 'web' should not be needed


doug2k1:
According to [Webpack docs](https://webpack.js.org/configuration/target/#target), 'web' is the default value for the target option. Both dev and prod configs has it set.

https://github.com/coryhouse/react-slingshot/blob/134b5f9b7dfcc3619a2af604e824f539b386ea47/webpack.config.prod.js#L21

kwelch:
I am not against the explicitness, but the link being broken is not so good.

coryhouse:
I just removed the outdated comment a8707054375b3f8c6ddd75f430e6156b75411299

Prefer leaving the setting to be explicit. Thanks!

PR #468, opened 2017-09-04
Update Webpack references to new docs

doug2k1:
I'm updating the Webpack references to the new docs (Webpack v2+), in both README.me and source code comments.

Note: I leave the lines that set `target: 'web'` on both Webpack configs untouched, as per #467 I think they could be removed.


kwelch:
Looks good, thanks for the updates.

PR #475, opened 2017-09-13
Removed redundant react-router dependency

ifeature:
As far as you're using react-router-dom package, it is no need, to include react-router dependency.

- In `router.js` you have an import reference to `IndexRoute`, which doesn't exist in react-router v4
- Moreover, react-router v4 designed in a way, when you can't describe all your existing routes in your app, especially, I'm talking about nested routes. So, you definitely don't need `routes.js`, because the routing part of component lies next to other component stuff, as you've specified in `App.js`
- `react-router` and `react-router-dom` libs have pretty the same functionality, but does it make sense to include both of them into project dependencies?


docs/FAQ.md
```diff
@@ -224,9 +224,8 @@ Nope. Redux is useful for applications with more complex data flows. If your app
 
 <a name="how-do-i-remove-react-router"></a>
 ### How do I remove React Router?
- 1. Uninstall React Router and routing related packages: `npm uninstall --save react-router`
- 2. Delete the following files: `src/routes.js`
- 3. Remove `import { Link, IndexLink } from 'react-router';` from top of `src/components/App.js`, add a reference to `src/components/FuelSavingsForm.js`, and replace body of (implicit) render with this: `<FuelSavingsPage />`.
+ 1. Uninstall React Router and routing related packages: `npm uninstall --save react-router-dom`
+ 3. Remove `import { Switch, NavLink, Route } from 'react-router-dom';` from top of `src/components/App.js`, add a reference to `src/components/FuelSavingsForm.js`, and replace body of (implicit) render with this: `<FuelSavingsPage />`.
```
kwelch: Could you also update the numbering?

kwelch:
Good catch! LGTM, after the one comment is addressed.

coryhouse:
Excellent point! I should have noticed those files were still lying around. Thanks for the cleanup! 👍

PR #481, opened 2017-10-04
Upgrade postcss-loader and remove old LoaderOptionsPlugin.

sheldonwjones:
I  ran into some problems when trying to updated the postcss-loader.

`
ERROR in ./node_modules/css-loader?sourceMap!./node_modules/postcss-loader/lib!./node_modules/sass-loader/lib/loader.js?sourceMap!./src/styles/styles.scss
Module build failed: Error: No PostCSS Config found in: ./react-slingshot/src/styles
`

I couldn't find a way to inline the postcss-loader options. This is probably because the LoaderOptionsPlugin was never meant to be used with webpack 3. https://webpack.js.org/plugins/loader-options-plugin/

This pull request solved my problems.

Thoughts?


webpack.config.dev.js
```diff
@@ -35,18 +34,6 @@ export default {
         collapseWhitespace: true
       },
       inject: true
-    }),
-    new webpack.LoaderOptionsPlugin({
-      minimize: false,
-      debug: true,
-      noInfo: true, // set to false to see a list of every file being bundled.
```
kwelch: Do we know what these options above do? Shouldn't we also bring other over.

kwelch:
Look good I put in a comment about the options that were unrelated to the `sass-loader`. Also, should the other loaders be updated to the `use` structure to stay consistent?

The build failure is due to `node@4` not being supported by a dependency of enzyme, but only due to our use of yarn.

webpack.config.dev.js
sheldonwjones: I'm pretty sure the minimize option is being handled correctly. It defaults to false and I set it to true for css in our production config.

The debug is being removed from webpack, and I couldn't find much on noInfo. Maybe it's an outdated option.

sheldonwjones:
We should probably update the other loaders to use the 'use' syntax. Has anyone tried converting them?

coryhouse:
Thanks @sheldonwjones - Yes, I agree moving everything to the 'use' syntax is a win. Willing to handle in this PR?

sheldonwjones:
I'll give it a shot.

sheldonwjones:
file-loader defaults to name: [hash].[ext]. I didn't set the name option and left it default. Is there a reason why we had it set to [name].[ext]?

https://github.com/webpack-contrib/file-loader

webpack.config.dev.js
kwelch: Awesome. Sounds good.

kwelch:
This looks good, thanks!

Issue #483, opened 2017-10-09
Demo App types errror


dawangdw:
Node version: v6.9.5

npm version: 3.10.10

Operating system: macOS

Command line used: npm start

Steps to reproduce: click Demo App in the home page, will see error in dev console - 
![image](https://user-images.githubusercontent.com/16933250/31323295-8e0d0612-ac6b-11e7-85f6-7d07604d0c02.png)

PR #484, opened 2017-10-09
Fix type error

dawangdw:
Related issue https://github.com/coryhouse/react-slingshot/issues/483


kwelch:
Look good. Thanks! 👍

coryhouse:
FYI - I had to roll this back since it broke the app. Please be sure to check before merging. :) I updated the propTypes to accept string or number. 

Thanks for the PR!

PR #486, opened 2017-10-09
use own store in the store spec instead of a generic redux store

ballPointPenguin:
It seems to me like the store spec should utilize the actual store code included in the project.

### DIFF
--- a/src/store/store.spec.js
+++ b/src/store/store.spec.js
@@ -1,12 +1,10 @@
 import * as ActionTypes from '../constants/actionTypes';
 
 import MockDate from 'mockdate';
-import { createStore } from 'redux';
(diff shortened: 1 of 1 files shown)


ballPointPenguin:
I would have expected coverage to increase as a result of acutally using the configureStore.js file. Is it possible that coverage is only calculated against files which are imported? If so then the drop in coverage % seen here might be the correct result of including configureStore.js into the coverage graph. The next step would be to add more testing to store.spec.js which hits more parts of configureStore.js.

coryhouse:
Good call on the change. And yes, I suspect coveralls is only considering test coverage on the files that are referenced by a spec file, so that's why it went down with this change. Thanks!

PR #480, opened 2017-10-01
React 16

kwelch:
This was the minimum I could do to update to React 16. 

The other PR included all packages updated as well. Once this is merged I will proceed with a package update across the board.

PR #488, opened 2017-10-19
Swap npm-run-all for concurrently

### LINKED ISSUE #120: Killing start script requires hitting Ctrl+C twice
Since migrating to eslint-watch, you must hit Ctrl+C twice to kill the start script. I [opened an issue with eslint-watch](https://github.com/rizowski/eslint-watch/issues/80) but if anyone has any suggestions, I'm all ears. 

https://github.com/rizowski/eslint-watch/issues/80

samit4me:
This should fix #120.

Have tested this on 
- `macOS 10.12.6` && `node 6.10.3` && `npm 5.4.2` && `yarn 1.2.1`.
- `windows 10 v1703` && `node 6.9.1` && `npm 3.10.8` && `yarn 1.2.1`.

Have also test that testing and linting work as expected (with and without errors) on both of the above.

Would be fantastic if others that have had issues in the past could pull this locally and test.


kwelch:
This builds, tests, and runs properly. It also resolves #120 as expected. 

I am good with it. 

@coryhouse any reason to not use `concurrently`, instead of `npm-run-all`

nickytonline:
@kwelch, CI is breaking though. I haven't had a chance to see why though.

kwelch:
It is actually a yarn issue. If a package has the engine field set yarn will not install on that version. It started back with the `react@16` merge. 

I would hold off on the merge until we can get some results from another Windows machine though (maybe?).

samit4me:
Looking at the changes in `package-lock.json` and `yarn.lock` it looks like they haven't been updated in sometime. In `master` autoprefixer is 7.1.4 listed in `package.json`, but it is 7.1.2 in `package-lock.json`.

Did you want me to roll back all the unrelated changes in `package-lock.json` and `yarn.lock` and try CI again?

samit4me:
The issue looks related to `rst-selector-parser` which was added into `yarn.lock` as a dependancy of enzyme 3

https://github.com/coryhouse/react-slingshot/pull/488/files#diff-8ee2343978836a779dc9f8d6b794c3b2R2308

samit4me:
Okay so I've looked into this a little bit more and it looks like Enzyme swapped out `scapel` for `rst-selector-parser` in [PR 1086](https://github.com/airbnb/enzyme/pull/1086/files#diff-640ddc435dee7ec652108f3a5d1083d4L42) and `rst-selector-parser` which requires node 5 or later [see here](https://github.com/aweary/rst-selector-parser/blame/master/package.json#L33).

This change looks like it was released with v3 of enzyme, so I looked into when this repo was updated to v3 and it seems this was done in #480. This is further confirmed with the CI failures in that PR, see [details](https://travis-ci.org/coryhouse/react-slingshot/jobs/281970610#L462).

Given that enzyme doesn't support node 4 and it is no longer active according to the [node release schedule](https://github.com/nodejs/Release) should/can this repo drop support for node 4? If so, let me know I'd be more than happy to help where possible.

coryhouse:
I'm open to dropping support for Node 4. 

Appreciate the PR! However, I'm hesitant to move to concurrently since its API is uglier.

samit4me:
Yeah agree, concurrently is uglier :)

Looks like there might be a fix for node 4 soon, see https://github.com/airbnb/enzyme/issues/1285

samit4me:
The CI should now pass as the upstream issue has been resolved. Is anyone able to re-run the failed ones?

kwelch:
I was able to restart Travis, but not AppVeyour

samit4me:
Thank you @kwelch, it looks like the lock files needed updating anyways to take advantage of the upstream fix. Looks like Travis CI is passing now 🎉 we can only hope appveyor does too when if finishes.

samit4me:
Everything seemed to pass, but `babel-node tools/testCi.js` appeared to timeout on node 7, very strange.

samit4me:
Is `concurrently` a no go? I totally agree it's much uglier than `run-npm-all` but I find this is easier to live with than needing to close the console when ctrl+c doesn't work. If this is a no go, might have to look into a different solution.

coryhouse:
I'm on Mac so I can't easily test at the moment, but if @nickytonline and @kwelch are good w/ the switch, I am so Windows people have a better experience

nickytonline:
I'm on a Mac only as well at the moment. Haven't had a chance to set up my IE and Edge VMs yet on the new machine.

lrn2prgrm:
I can confirm the issue in windows 10. currently doing: 
> netstat -ano | findstr :PORT
> taskkill /PID PID_NUMBER /F

coryhouse:
Merged. Thanks for the PR!

coryhouse:
I just reverted this commit to resolve #500

coryhouse:
Sorry, my mistake, the actual issue was caused by #497. I just re-implemented this PR. Thanks again!

PR #495, opened 2017-10-26
Use local binaries

techmexdev:
This prevents global & local version conflicts, and doesn't require users to globally install the binaries in their machine.


kwelch:
I am pretty sure we have this before and we moved away from it. @coryhouse Do you recall why that was?

coryhouse:
One need not install global - any packages referenced in an npm script are added to the path. And I understand resolution starts at local node_modules, so I see no reason to make this change.

PR #497, opened 2017-10-26
Update package.json

techmexdev:


coryhouse:
Good call. This makes sense to me. @nickytonline @kwelch any concerns?

coryhouse:
Just removed extra dashes in command to resolve bug.

PR #501, opened 2017-11-03
Removed double hyphen from "jest -- --watch"

plneto:
I was getting the following output when running "npm start -s" and the app wasn't starting. I was only able to run it after changing it to "jest --watch".

```
Starting app in dev mode...
No tests found
In /Users/paulo/Development/react-slingshot
  50 files checked.
  testMatch: **/__tests__/**/*.js?(x),**/?(*.)(spec|test).js?(x) - 12 matches
  testPathIgnorePatterns: /node_modules/ - 50 matches
Pattern: "--watch" - 0 matches
```


coryhouse:
I just fixed this in a separate commit, but thanks for reporting!

PR #504, opened 2017-11-13
Help with jest testing -- Gender select form with jest

XanderM1960:
Hello, 

I took the Pluralsight course of yours which was great. This slingshot is very helpful.

This is a strange way to ask for help, but could you look at the failing test in this pull request and suggest how GenderSelectPage.handleSubmit can work normally on the dom (jsdom) during the test?

Everything works fine in the browser of course.

(you of course don't want to merge this ... )


kwelch:
This is one of the best ways I have seen ask for help on something. This is much simplier to follow parse and run than just a description in a ticket. Thank you for that. 

I will do my best to check this tonight.

kwelch:
I put in a PR with a recommend change to your approach instead of using a selector to get the checked item. Instead using local state to keep and maintain value. I am have not done anything with your approach, so it was difficult to find next steps.

XanderM1960:
Thanks. I will take a look.

PR #507, opened 2017-11-22
Stop excluding node_modules for styles

### LINKED ISSUE #506: Styles in npm packages can't be imported
We're ignoring npm packages in the style loader webpack config. Need to stop ignoring so styles in npm packages can be imported.

coryhouse:
Someone pointed out they couldn't import styles from npm packages because we're excluding node_modules. This PR fixes #506

### DIFF
--- a/webpack.config.dev.js
+++ b/webpack.config.dev.js
@@ -96,7 +96,6 @@ export default {
       },
       {
         test: /(\.css|\.scss|\.sass)$/,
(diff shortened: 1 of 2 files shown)


kwelch:
Seems simple enough and not sure how that would break any css since we are not using CSSModules. That is the only time I have needed to be specific with include and excludes.

Issue #511, opened 2017-12-10
implementing less loader


sag1v:
Hi Cory,
In my team we are working with `less` instead of `sass` and i often add the `less` loader to the project.  
Do you think it is OK to add the `less`  loader to webpack, so slingshot will support `less` out of the box? 
Maybe after or before the `sass` loader in the pipe.

coryhouse:
Yeah, I can see your point. I'm open to a PR. @nickytonline @kwelch thoughts?

nickytonline:
@coryhouse, I don't see any issues with supporting it OOTB.

kwelch:
Sounds like a great idea! 👍

sag1v:
Great. 
here you go #512

nickytonline:
I forgot that we added the `package-lock.json` a while back.

kwelch:
Merged.

PR #512, opened 2017-12-11
added less-loader support

sag1v:
per #511


sag1v:
@kwelch I thought the same, though i got an error when `less-loader` was last in the pipe

PR #519, opened 2017-12-23
Add blog post showing how to deploy to Heroku

guilatrova:
Not sure if it was supposed to be so simple haha :joy:

Maybe I need do add more details or create a specific section for this?


kwelch:
I think this works we can adjust format as we have other methods of deployment. 

What do you think @nickytonline and @coryhouse?

coryhouse:
Thanks! 👍

PR #522, opened 2017-12-27
Adding docker support

dockerizeallthethings:
- Adding docker image files for both prod and dev use
- Adding instructions to top level README.md to:
  - Run a copy of the prod version hosted on [hub.docker](https://hub.docker.com/r/allthethings/react-slingshot/) to try out the project
  - Build and run an image that will listen to changes on the host machine (to allow developers to use an IDE on their host machine)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM node:8-alpine
+
+WORKDIR /code/
+
+EXPOSE 4000 4001
+
+COPY . .
+
+RUN npm run setup
+
+ENTRYPOINT npm run build
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,12 @@ FROM node:8-alpine
 
 WORKDIR /code/
 
-EXPOSE 4000 4001
+EXPOSE 3000 3001
 
 COPY . .
 
+RUN npm install --unsafe-perm
+
 RUN npm run setup
 
-ENTRYPOINT npm run build
+ENTRYPOINT npm start
--- a/Dockerfile-prod
+++ b/Dockerfile-prod
@@ -1,0 +1,11 @@
+FROM node:8-alpine
+
+WORKDIR /code/
+
+EXPOSE 4000 4001
+
+COPY . .
+
+RUN npm run setup
+
+ENTRYPOINT npm run build
--- a/README.md
+++ b/README.md
@@ -45,8 +45,18 @@ This will run the automated build process, start up a webserver, and open the ap
 * **Install C++ Compiler**. Browser-sync requires a C++ compiler on Windows. [Visual Studio Express](https://www.visualstudio.com/en-US/products/visual-studio-express-vs) comes bundled with a free C++ compiler. Or, if you already have Visual Studio installed: Open Visual Studio and go to File -> New -> Project -> Visual C++ -> Install Visual C++ Tools for Windows Desktop. The C++ compiler is used to compile browser-sync (and perhaps other Node modules).
 
 **Try It Out With [Docker](https://www.docker.com/):**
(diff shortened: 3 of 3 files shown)


nickytonline:
@dockerizeallthethings, thanks for the PR! Could you open an issue and link it to this PR? In general, we like to have a discussion about what goes into the project and an issue is the perfect place to do that. Cheers.

coryhouse:
Thanks for the PR! Dumb question: How would we update the Docker image hosted on hub.docker later?

dockerizeallthethings:
If you have accounts on hub.docker I should be able to add them as
contributors.

Also, you're certainly welcome to host your own version on hub.docker, if
you'd prefer.

Lemme know if you have any other questions, and feel free to send me some
hub.docker usernames to add :).

--RG

coryhouse:
Great! I'm `housecor`

dockerizeallthethings:
Done!

I also created an issue linked to the PR, as per an request from a Nick
Taylor (to which I can't seem to directly respond).

Happy New Year!

--RG

kwelch:
I don't know much about docker so I cannot approve or deny of these changes. I am definitely excited about them though.

mikedevita:
I actually do something very similar to this already for a few projects, it helps to alleviate the complexities with managing various versions of nodejs and npm packages on my mac. I mount the root of the project folder into the container so i can then run the build if i wanted.

nickytonline:
This is for implementing #529.

nickytonline:
I haven't tried this locally yet, will try to take a peek at this tomorrow. Just curious why alpine instead of slim for Node? I haven't done a whole lot of Docker yet. New job's immersing me though 😉

coryhouse:
I also don't know Docker well enough yet to vet this. I'm hoping to dive in soon for work but for now I'm no help on this.

ikim23:
Hi!

At first good work @dockerizeallthethings. I wanted to do this myself, but you were faster than me :).

Couple notes from my side:
* HMR is not working on Docker for Windows, because of the way how volume binding works. To make it work, you must add poll: true to watchOptions inside webpackDevMiddleware config
* This is just my thing, but I personally prefer to use docker-compose over docker cli. Even though in this case we really do not need to orchestrate multiple containers, I find it much more easy to work with. E.g.: If you have one service inside docker compose with all necessary configurations (volume/port binding, container name, tty, ...) you can than just type `docker-compose up hmr` instead of `docker run -v "$(pwd)"/src:/code/src -p 3000:3000 -p 3001:3001 --name react-slingshot -d react-slingshot`
* During dev, I would probably rather use full node container instead of alpine. Sometimes I find it useful to connect inside the container and execute bunch of commands from console, but again this is just a personal thing...

I can contribute with docker-compose setup if you are interested.
Please, let me know.

nickytonline:
@dockerizeallthethings, sorry that this PR and issue has gone stale. Any ideas for the first point that @mikedevita mentions?

acmitch:
I'm not a huge fan of this for a couple of reasons:
  - Only one Dockerfile should be provided and it should closely mimic a PROD environment. 
  - The PROD dockerfile it should not build the project. This causes a lot of ENV headache. The dist folder should be built like normal, then copied over to the Docker image on creation.

If someone wants to develop using Docker, that should be put back on the developer, this PR is making A LOT of assumptions. 

For example, I use react-slingshot in a multi-project build, but my react-slingshot project has dependencies on other projects that have to be built first. Therefore, the Dockerfile has to change accordingly.

In my opintion, this is a good start but not ready to merge. I'm going to FORK react-slingshot this weekend and try to implement some ENV configuration changes and I'll integrate Docker as well.

acmitch:
If anyone is curious about my Docker comment, check out the PR: https://github.com/coryhouse/react-slingshot/pull/590

PR #525, opened 2017-12-30
Remove references  .eslintrc and .babelrc from FAQ and readme.

vitalikpi:
Both .eslintrc and .babelrc were move into package.json. As a newcomer I was confused why they are absent.


coryhouse:
Good catch. Thanks! 👍

Issue #526, opened 2017-12-30
Sass calc() doesn't work properly in this kit


leonimurilo:
I have the following sass code for a grid column:
Working fine in Codepen: https://codepen.io/anon/pen/zpwpOm?editors=1100
```sass
// 1rem = 10px
$grid-width: 114rem
$gutter-horizontal: 6rem
.row
  max-width: $grid-width
  .col-1-of-2
    width: calc((100% - #{$gutter-horizontal}) / 2)
```

```jsx
<div className="row">
  <div className="col-1-of-2">Col 1 of 2</div>
  <div className="col-1-of-2">Col 1 of 2</div>
</div>
```

Since the grid width is 1140px (max size triggered) and $gutter-horizonta is 60px, each column width should be 540px so I would have 540px + 60px + 540px = 1140px. However for some reason the column width that appears when I inspect the element in any browser is 47% which is no sense since 47% is 535.8px not 540px.

Why is the following code being converted to no sense percentages?
`    width: calc((100% - #{$gutter-horizontal}) / 2)`

It should be converted to 540px not 47%
![screen shot 2017-12-30 at 10 46 31](https://user-images.githubusercontent.com/14261943/34454352-c4d01c86-ed50-11e7-91b3-4b410c50ca02.png)

Of course I end with a unexpected spacement aftet the second column:

![screen shot 2017-12-30 at 10 07 22](https://user-images.githubusercontent.com/14261943/34454351-c4ae222a-ed50-11e7-834f-e28cf88de23c.png)

For some reason the width in the compiled css is 47% while in any other place I test this code, like in a separate node-sass project or in Codepen it works perfectly.

node-sass: 4.7.2
sass-loader: 6.0.6

Node version:
6.2.2 (also tried 8.9)
npm version:
3.9.5
Operating system:
macOS High Sierra
Command line used:
npm start

nickytonline:
@leonimurilo,

If everything builds and the SASS is compiled to CSS, I don't see how there is a problem with React Slingshot and I find it highly unlikely that there is an issue with node-sass or sass-loader.

Do you have a public repo to compare your SASS with that of the Code Pen to further investigate the issue? Thanks.

leonimurilo:
@nickytonline,

Yes I do: https://github.com/leonimurilo/Natours-React
I am using the 7-1 sass pattern and the code that I showed in codepen is distributed in:

### sass files
./src/styles/abstracts/_variables.sass
[...]

PR #528, opened 2018-01-01
Remove less support due to #526

leonimurilo:
Remove less support as an alternative to issue #526


coryhouse:
Looks good. @sag1v - This rolls back your less add since it caused a regression. Would welcome a PR from you as an alternative to removing the less support. Suggestion: Need to declare Less loader separately from Sass so the same loaders don't apply to both.

sag1v:
@coryhouse  fair enough.
@kwelch [i guess the ordering does matter](https://github.com/coryhouse/react-slingshot/pull/512#pullrequestreview-82548827) 😞

nickytonline:
Everything looks good, when comparing it to the addition of LESS in #512.

coryhouse:
Merged. Thanks for reporting!

PR #539, opened 2018-02-10
Lock the node version to 8.

coryhouse:
Found a handy trick that should help enforce consistency: https://twitter.com/housecor/status/962347301456015360

Look good @nickytonline @kwelch ?


coryhouse:
Ah, good point on dev dep. Given that Node is only for dev purposes here, that is the better choice.

The bigger problem: Appveyor ain't happy about this change because it appears the specified version isn't available on their platform. 🤔

nickytonline:
That's odd that they don't have it given that Node 8.9.x is the current LTS. I'm totally OK with going all in on Node 8 as it's only for building the front-end. I can't think of any negatives.

coryhouse:
I've opened an issue with Appveyor.

nickytonline:
So looks like we're good to go as soon as the AppVeyor issue is sorted out.

nickytonline:
Looks like you got AppVeyor sorted out @coryhouse. 👍

nickytonline:
Nevermind. My eyes were deceiving me. It's just Coveralls.

kwelch:
Why are we dropping lock files?

coryhouse:
Just added the lock files back. Finally got the CI quirks resolved so we should be good now.

kwelch:
Awesome. Excited to have this locked in! Nice work

Issue #407, opened 2017-04-05
Update ESLint Config


kwelch:
After the changes made in #405, there is a lot of eslint config now in the `package.json`. 

In effort to clean this up, there was a [suggestion to move Slingshot to a baseline preset](https://github.com/coryhouse/react-slingshot/pull/405#issuecomment-291501749). 

This issue is to capture that ask and open the discussion to others about what the eslint config preset should be set to. 

Options
- ESLint Recommended Baseline
  - Do we include other recommended baselines (ie. react, import, etc.)
- ESList shared preset
  - Which preset?
    - AirBnB - No  - too opinoinated see #35 
    - CRA?
    - Standard/Standard-Semi?
- Package Slingshot Config
  - Is the current config at a place that is right sized and can be shared as a packaged config

coryhouse:
Thanks @kwelch! I vote for ESLint recommended. If you prefer something else, enter a comment so people can upvote via a 👍

kwelch:
I would not mind seeing us incorporate [prettier](https://github.com/prettier/prettier) (or [prettier-eslint](https://github.com/prettier/prettier-eslint)) to have a simple consistent format and use baseline recommended for best practice / bug catching type linting.

coryhouse:
I like Prettier, but also recognize it's a polarizing idea - As an anecdote, in our office, a significant portion of people find the idea of having their code changed automatically in any way is unsavory. Many prefer to "learn from their mistakes" by manually fixing what ESLint reports. I can appreciate that. And there are certainly things Prettier does that I don't like - such as splitting long imports into multiple lines. So I feel ESLint recommended is the most populous friendly approach at the moment.

kwelch:
I understand that. It also adds a bit more weight than I think we would stand to institute and support in a boilerplate. It is a better fit for library.

kwelch:
I think we are solid on the approach. Adding a help wanted and good first PR label.
😄

qborreda:
@coryhouse - you could have both, ESlint would still throw errors and warnings at you via your editor, while prettier can make sure the output of the file is exactly the same no matter who has written it.

coryhouse:
@qborreda True, though it seems the recommended approach is to disable any formatting related warnings from ESLint to avoid the duplication. It would seem noisy to warn someone about something that will be automatically fixed on save/commit. I think that's the sales pitch for prettier - it makes coding more luxurious because you literally stop thinking about formatting altogether.

[...]

PR #540, opened 2018-02-12
Update ESLint Config #407

Mottoweb:
Hey,

I stumbled upon #407, so i thought i could give it a go for a first good PR.

Let me know if i missed anything.
Things were done:
1. Remove anything lint related from package.json
2. Create .eslintrc.json with same rules.
3. Remove custom rules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,7 +171,8 @@
     "extends": [
       "eslint:recommended",
       "plugin:import/errors",
-      "plugin:import/warnings"
+      "plugin:import/warnings",
+      "plugin:react/recommended"
     ],
     "plugins": [
       "react"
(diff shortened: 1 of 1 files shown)


.eslintrc.json
```diff
@@ -0,0 +1,28 @@
+{
```
kwelch: I believe we still want the config to live in `package.json`. This was done intentionally to reduce errors when the project was copied. See #405

.eslintrc.json
```diff
@@ -0,0 +1,28 @@
...
+  "root": true,
+  "extends": [
+    "eslint:recommended",
+    "plugin:import/errors",
+    "plugin:import/warnings"
```
kwelch: I would expect the import plugin to be listed in plugs. Is it inherited by another?

.eslintrc.json
```diff
@@ -0,0 +1,28 @@
...
+  ],
+  "plugins": [
+    "react"
```
kwelch: I like that we are importing this, I do believe we should find a export preset of rules we can extend here to cover (most or all) of the rules we had previously set.

kwelch:
Thanks, I have put in a few comments. I appreciate you taking a stab at this. :+1:

.eslintrc.json
Mottoweb: Then what is the point of removing custom rules and creating dependency on other packages? The very last comment section of the issue #407 suggests a PR that removes custom rules, imports default, or am i wrong?

.eslintrc.json
Mottoweb: Ok no prob)

.eslintrc.json
Mottoweb: Could not trace where exactly it is inherited from but, it is.

kwelch:
@coryhouse & @nickytonline What do you think? I still want to pull it down and run lint a few times and fix to see if it causes any changes.

nickytonline:
I'm OK with you testing it locally to see if any other changes are required. It's using recommended/defaults for linting, so unless you see something glaring after you test it, I'd say we're probably good to go if Cory's OK with the changes.

coryhouse:
Linting runs clean for me so PR looks 👍

nickytonline:
Thanks for the PR @Mottoweb!

Mottoweb:
Glad to be of help!

Issue #545, opened 2018-03-10
npm run remove-demo removes styles folder


nharrisanalyst:
Why is that?
As a developer it is confusing where to put SASS or css files should we keep the styles folder to be clear where styles should go?

I know that in create-react-app it is very clear where styles should go.

coryhouse:
I see your point. I'm open to a PR that leaves the styles folder.

nharrisanalyst:
this has been fixed

PR #546, opened 2018-03-13
keeps style folder but deletes everything in it

nharrisanalyst:
Changes the tools/removDemo.js file so that the styles folder is not deleted but that everything in the styles folder is deleted


nharrisanalyst:
per https://github.com/coryhouse/react-slingshot/issues/545#issue-304096397

PR #549, opened 2018-03-26
Fix prop types silently being evaluated to nothing

vinczemarton:
Problems currently

1. `oneOf` is a union of possible values not types
2. indexing `oneOf` with `number, string` instead of calling it as a function is not a syntax error but `oneOf` has no property called the string value of `string` so the prop type will most likely be undefined which will silently not catch type mismatches.


coryhouse:
Great catch. Thanks! 👍

PR #561, opened 2018-04-29
Upgrade to webpack 4. Add HardSourceWebpackPlugin for performance

innoist:
-- upgraded packages to support webpack
-- extract-text-webpack-plugin needs to be at its next version to support
-- added `hard-source-webpack-plugin` to cache the build of node modules so successive dev builts are fast.
-- all tests are passing
-- webpack 4 has now its own contenthash variable so renamed it to : [md5:contenthash:hex:20]
https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/763


webpack.config.dev.js
```diff
@@ -1,6 +1,7 @@
 import webpack from 'webpack';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import path from 'path';
+var HardSourceWebpackPlugin = require('hard-source-webpack-plugin');
```
coryhouse: Is there a reason you used var instead of import?

innoist:
PR updated. Thanks

coryhouse:
Thanks so much for your hard work on this! 👍

Issue #594, opened 2018-11-01
Wrong absolute path in configs - SASS styles


vbatrla:
I have found wrong path in file **webpack.config.dev.js** and **webpack.config.prod.js** in plugin sass-loader. Take a look at line 116 of this file. The problem should be in includePath parameter. I'm guessing there should be option "styles" in third parameter instead of "scss". If I'm wrong just let me know why. Thank you.

```
{
  loader: 'sass-loader',
  options: {
	includePaths: [path.resolve(__dirname, 'src', 'scss')],
	sourceMap: true
}      
```

coryhouse:
Ah, good point. I think it makes sense to omit the last arg and just point it at src. Would you be willing to do a PR?

kwelch:
Closed in #615

PR #595, opened 2018-11-02
Fixing invalid SASS-Loader paths

vbatrla:
I have found wrong path in file webpack.config.dev.js and webpack.config.prod.js in plugin sass-loader. Take a look at line 116 of this file. The problem should be in includePath parameter. I'm guessing there should be option "styles" in third parameter instead of "scss". If I'm wrong just let me know why. Thank you.

```
{
  loader: 'sass-loader',
  options: {
	includePaths: [path.resolve(__dirname, 'src', 'scss')],
	sourceMap: true
}      
```

Issue: https://github.com/coryhouse/react-slingshot/issues/594

### DIFF
--- a/webpack.config.dev.js
+++ b/webpack.config.dev.js
@@ -113,7 +113,7 @@ export default {
           }, {
             loader: 'sass-loader',
             options: {
-              includePaths: [path.resolve(__dirname, 'src', 'scss')],
+              includePaths: [path.resolve(__dirname, 'src', 'styles')],
               sourceMap: true
             }
           }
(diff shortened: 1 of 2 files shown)


coryhouse:
Thanks for the PR! I'm thinking we can be less specific and point it simply at the src directory so people can place Sass anywhere in there. Thoughts?

vbatrla:
I've seen your email but couldn't respond earlier. I think there should be
one configuration file where could be placed all specific paths and
variables. So user could have a possibility to set custom paths and
variables.

pá 2. 11. 2018 v 12:43 odesílatel Cory House <notifications@github.com>
napsal:

nickytonline:
Just following up on this PR. Pointing to the `src` directory makes sense to me, then there's nothing else to configure.

nickytonline:
Just following up on this again @vbatrla.

PR #613, opened 2019-04-28
Added node 11 to CI

nickytonline:
Update CI to support Node 11 ~and 12~.


kwelch:
Can you add why `12` didn't work so we can look at what changes are necessary to support it.

nickytonline:
Yeah, will do. It's on my TODO. ;)

PR #620, opened 2019-07-07
Add viewport meta tag

alepacheco:
Adding this tag makes it look ok on mobile


coryhouse:
Great idea. Thanks! 👍

PR #621, opened 2019-07-07
Fix async await

alepacheco:
Async/Await does not work out of the box, this fixes it

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1548,6 +1548,26 @@
         "@babel/helper-plugin-utils": "^7.0.0"
       }
     },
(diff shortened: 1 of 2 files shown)


nickytonline:
Not related to this PR, but one thing I noticed going through this PR is that production has presets set to ie9 and other environments have none set. I think we could revise this list to be a little more up to date. People can always alter the browserlist settings after. Thoughts @housecor?

coryhouse:
Looks great. Thanks! 

And agreed @nickytonline - I'd welcome a PR

PR #639, opened 2019-12-10
add link to replit run

eankeen:
# Pull Request Template

i think it would be pretty cool to let people see how this react slingshot boilerplate works (and edit / preview the program) right in their browser, without any manual downloads or installation. i'm thinking we can add a 'run on repl.it' button, which redirects to something like [this](https://repl.it/@eankeen/run-react-slingshot):

image

i suppose if this is what we want, then we could possibly want the url to be owned by a more agnostic entity

## Checklist

- [x] Latest code from master has been merged into the pull request branch
- [x] Code is camelCased
- [x] No commented out code (if required, place // TODO above with explanation)
- [x] No linting issues
- [x] Automated tests exist and pass
- [x] Build is successful (`npm run build`)
- [x] Works in IE 11, Chrome, Firefox, and Edge

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,2 @@
+language = "nodejs"
+run = "npm start"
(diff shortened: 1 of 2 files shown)


coryhouse:
Yeah, I can see the benefit, but would like to avoid adding another piece to maintain to the project. Thanks for the idea though!

